PR #3, opened 2018-04-27
Use interface index instead of name in libconfig

imanti:
This patch allows lldpad to use any network interface name.

Right now lldpad does not accept interface names containing non
alphanumeric characters (except '_' and '-'). As Markus Mehlan reported
(http://lists.open-lldp.org/pipermail/lldp-devel/2017-March/001251.html)
this is a constraint originally from libconfig.

Libconfig manual states that:
*http://www.hyperrealm.com/libconfig/libconfig_manual.html

"All names are case-sensitive. They may consist only of
alphanumeric characters, dashes (‘-’), underscores (‘_’), and
asterisks (‘*’), and must begin with aletter or asterisk. No
other characters are allowed."

Given this restriction we thought the best idea would be to interact
with libconfig using a "sanitized" version of the interface name. The
names in this patch are generated using the index of the interface as
returned by if_nametoindex() function. The name string has the format
"ifX" where X is the interface index.

### DIFF
--- a/config.c
+++ b/config.c
@@ -56,6 +56,21 @@
 config_t lldpad_cfg;
 
 /*
+ * config_ifkey - Generates a config key
+ *
+ * Given an interface name this functions generates
+ * a key (based on interface's index) suitable
+ * to pass to libconfig.
+ *
+ */
+void config_ifkey(const char *name, char *ifkey) {
+	int index = if_nametoindex(name);
+	
+	if(index)
+		sprintf(ifkey, "if%d", index);
+}
+
+/*
  * init_cfg - initialze the global lldpad_cfg via config_init
  *
  * Returns true (1) for succes and false (0) for failed
@@ -451,14 +466,15 @@ static int lookup_config_value(char *path, union cfg_get v, int type)
 int get_config_setting(const char *ifname, int agenttype, char *path,
 		       union cfg_get v, int type)
 {
-	char p[1024];
+	char p[1024], ifkey[IFNAMSIZ];
 	int rval = CONFIG_FALSE;
 	const char *section = agent_type2section(agenttype);
 
 	/* look for setting in section->ifname area first */
 	if (ifname) {
+		config_ifkey(ifname, ifkey);
 		snprintf(p, sizeof(p), "%s.%s.%s",
-			 section, ifname, path);
+			 section, ifkey, path);
 		rval = lookup_config_value(p, v, type);
 	}
 
@@ -475,15 +491,16 @@ int get_config_setting(const char *ifname, int agenttype, char *path,
 int remove_config_setting(const char *ifname, int agenttype, char *parent,
 			  char *name)
 {
-	char p[1024];
+	char p[1024], ifkey[IFNAMSIZ];
 	int rval = CONFIG_FALSE;
 	config_setting_t *setting = NULL;
 	const char *section = agent_type2section(agenttype);
 
 	/* look for setting in section->ifname area first */
-	if (ifname) {
+	if (ifname) { 
+		config_ifkey(ifname, ifkey);
 		snprintf(p, sizeof(p), "%s.%s.%s",
-			 section, ifname, parent);
+			 section, ifkey, parent);
 		setting = config_lookup(&lldpad_cfg, p);
 	}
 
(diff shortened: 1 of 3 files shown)


praeluceo:
Note from our validation:

# Testing of PATCH "Use interface index instead of name in libconfig"
## Standard interface names with patch applied

First, I confirmed that lldpad/lldptool work with interfaces with no "." in the name. The basic commands I tested:

```
lldptool -ti ens7
lldptool -tni ens7
lldptool set-tlv -i ens7 -V sysName -c enableTX=no
lldptool set-tlv -i ens7 -V sysName -c enableTX=yes
```

Inside lldpad.conf, I can see the data structure for the interface MIBs--here is the lldp section:

```
lldp : 
{
  if2 : 
  {
    tlvid00000001 : 
    {
      info = "045254007BAED4";
    };
    tlvid00000002 : 
    {
      info = "035254007BAED4";
    };
  };
  if3 : 
  {
    tlvid00000001 : 
    {
      info = "0490E2BAC63241";
    };
    tlvid00000002 : 
    {
      info = "0390E2BAC63241";
    };
    tlvid001b2101 : 
    {
      enableTx = true;
    };
    tlvid001b2102 : 
    {
      enableTx = true;
    };
    adminStatus = 3;
    tlvid00000005 : 
    {
      enableTx = false;
    };
```

So ens7 is mapped to if3 in lldpad.conf.

As I toggle TLVs (like sysName w/ tlvid tlvid00000005) the enableTx for the corresponding section changes to true/false. This matches the behavior on origin/master. The only difference I can see is that previously lldpad.conf contained the actual interface name (ens7), but now--since the patch maps real interface names to indexes provided by if_nametoindex(), the interface index is printed to lldpad.conf. It appears like this doesn't have an impact on lldpad/lldptool usage, but if there are any tools out there that scrape/read this file they'll need to do some mapping.

## Summary of patched openlldp w/ standard (ens7) interface names

I can spot no obvious regressions in functionality when using standard interface names. The only difference is the format of lldpad.conf using the index for
all interfaces instead of the real name as it is on origin/master.
"." interface names with patch applied

First, I tested some basic commands:

```
lldptool -ti ens.7
lldptool -tni ens.7
lldptool set-tlv -i ens.7 -V sysName -c enableTX=no
lldptool set-tlv -i ens.7 -V sysName -c enableTX=yes
```

All the commands ran fine. I then inspected lldpad.conf and saw that there is a reference to device "if3" which contains the dcbx/lldp info for
"ens.7". Again, this is due to the mapping provided by the patch. When I toggle enableTx on tlvs and inspect lldpad.conf, I can see the enableTx=
line change. Looks good.

## Summary of patched openlldp w/ non-standard (ens.7) interface names

I can spot no obvious differences in behavior when "." is and is not in the name. The patch does add functionality.
Summary and todo

So strictly lldp-wise I think the patch is fine.

I'd like to further test dcbx features, but they will take slightly longer. I'm getting a b2b config setup where I can do on-the-wire validation of the lldp packets to confirm that local configurations changes are being applied. I'll try to get these results soon but so far so good.

Original feature request: http://lists.open-lldp.org/pipermail/lldp-devel/2017-March/001251.html
Clarification from author: http://lists.open-lldp.org/pipermail/lldp-devel/2018-January/001278.html
Patch: http://patchwork.open-lldp.org/patch/298/

praeluceo:
Validator tested DCBx with NNT in a back-to-back configuration and with a switch. Willing and non-willing where applicable. In both cases, interfaces with "." in the name worked the same as they did pre-patch.

It looks fine to us, approving PR.

orgcandman:
There is an issue with this patch if the hardware module load order changes, or if hardware config changes are made, or even if there are races on the PCI bus.  The ifindex can change boot to boot (and it has been reported, ex: https://bugzilla.redhat.com/show_bug.cgi?id=1997064), so this patch will cause config changes to be lost, and worse - misapplied.

I am asking if we can revert this patch for the itme being.
@imanti and @dmertman ?

orgcandman:
Also, @praeluceo

praeluceo:
@dmertman  and I spoke about this, this week. We agree that this issue is definitely a bug and the predictable network interface naming service will cause it to come up wherever we use names.

Our conversation lead us to believe we likely should change the behavior such that the MAC addresses for the interfaces are stored in the config, with the exclusion of locally administered addresses. When lldpad starts up, it satisfies configs based on MAC, ignoring LAA interfaces, leaving those with a default queue map.

What do you think? We agreed that reverting  the patch makes the most sense for now since customers are actively running up against this.

PR #4, opened 2018-05-24
memleak on received TLVs from modules

cleech:
Most of the TLV modules that have an rchange handler for received TLVs
seem to get the return values wrong, returning 0 or TLV_OK without
freeing or storing the unpacked TLV to be freed later. That leaks the
allocation, as rxProcessFrame believes the module has claimed ownership.

In a test setup, it's probably easiest to see by enabling some TLV type
on one side of a connection only.  Or, any unexpected TLV that doesn't
get handled will be erroneously leaked by the EVB modules.

### DIFF
--- a/lldp_8021qaz.c
+++ b/lldp_8021qaz.c
@@ -1924,7 +1924,7 @@ int ieee8021qaz_rchange(struct port *port, struct lldp_agent *agent,
 	struct ieee8021qaz_unpkd_tlvs *rx;
 
 	if (agent->type != NEAREST_BRIDGE)
-		return 0;
+		return SUBTYPE_INVALID;
 
 	qaz_tlvs = ieee8021qaz_data(port->ifname);
 	if (!qaz_tlvs)
@@ -2005,7 +2005,7 @@ int ieee8021qaz_rchange(struct port *port, struct lldp_agent *agent,
 		}
 	}
 
-	return TLV_OK;
+	return SUBTYPE_INVALID;
 }
 
 static void ieee8021qaz_free_rx(struct ieee8021qaz_unpkd_tlvs *rx)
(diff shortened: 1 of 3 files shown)


praeluceo:
We are validating this patch today, thank you for submitting it and your patience! I hope to have an update this evening or tomorrow morning.

praeluceo:
@cleech Do you have an example configuration that experiences memleaks that we could configure our test against?

praeluceo:
Validation of this PR showed no conflicts, however no memory leak was detected during testing either before (without the patch) or after lldpad had been patched. Validation OKs this patch, @dmertman can provide development's position on it the PR.

cleech:
Hi, sorry about that.  The reports that I was dealing with were caused by running lldpad in a VM under VMWare ESXi, I don't know what type of TLVs were being received. But it looks like any TLV that doesn't have a receive handler should work.

I just tried enabling LLDP-MED on both ends of a back-to-back connection (actually two VMs connected to an OVS vswitch with flow rules to forward LLDP frames between them instead of stopping at the switch...)

I ran "valgrind --leak-check=full lldpad" and set things up with this script on both ends

  #!/bin/bash
  NET=ens10

  lldptool set-lldp -i "$NET" adminStatus=rxtx
  lldptool -T -i "$NET" -V sysName enableTx=yes
  lldptool -T -i "$NET" -V portDesc enableTx=yes
  lldptool -T -i "$NET" -V sysDesc enableTx=yes
  lldptool -T -i "$NET" -V sysCap enableTx=yes
  lldptool -T -i "$NET" -V mngAddr enableTx=yes
  lldptool -T -i "$NET" -V LLDP-MED enableTx=yes devtype=class1

Let it sit for a while, then terminate the lldpad process and see multiple lost malloc blocks from a call chain that goes through process_rx_frame > rxProcessFrame > create_tlv

The longer it runs, the more lost blocks from create_tlv pile up

PR #5, opened 2018-06-06
DCB set state operation return value check

ShayAgros:
Added a return value check for 'set_dcb_state'. The function
didn't check whether the 'set_hw_state' was successful.
Without this check 'dcb sc <if> dcb [on/off]' returns 'Successful'
even when operation is not supported in driver.

Fixes: 527ac435e46c ("lldpad: Do not enable CEE mode if explicitly disabled")
Signed-off-by: Shay Agroskin <shayag@mellanox.com>


praeluceo:
Thanks for forwarding this to a PR in Github, we have it under test now and will reply when we have results (likely next week).

PR #6, opened 2018-06-18
ecp22: deinit l2_packet_data before freeing ecp22

sregister:
When a link is shut, an ecp22 and l2_packet_data structure is removed for
the corresponding interface. l2_packet_data has a socket that needs to be
closed via l2_packet_deinit before the ecp22 structure can be freed.

Failure to close the socket causes a segfault once lldpad has exceeded the
maximum open file descriptors.


praeluceo:
This has been validated as mergable, assigning to Dave to merge.

PR #8, opened 2018-07-05
Fix application tlv display

imanti:
According to 802.1qaz standard the length of the Application
Priority Table is stored in a 9 bit variable. Current function
ieee8021qaz_print_app_tlv uses an 8 bit integer to iterate over
the table. This commit changes the iterator to a 16 bit integer.


praeluceo:
Validation shows that this compiles fine with no obvious regressions. Assigning to Dave to merge.

PR #9, opened 2018-07-09
lldpad: Support DSCP selectors in APP TLV's

pmachata:
The P802.1Qcd/D2.1 standard draft introduces a new APP TLV: DSCP, with
selector value of 5. Don't reject APP TLV's with selector 5, and
sanitize the PID value to not be out of bounds for DSCP.

Signed-off-by: Petr Machata <petrm@mellanox.com>


praeluceo:
We will get this under test, thank you for your PR!

PR #10, opened 2018-08-30
Fix gcc <v7.1 Wimplicit-fallthrough error

imanti:
The __attribute__((fallthrough)); (used to allow falling through
cases in a switch statement) is not defined before gcc version 7.1.
To fix this there is a marker comment (see details below) so this
commit replaces the attributes by the marker comment.

Documentation for Wimplicit-fallthrough:
https://gcc.gnu.org/onlinedocs/gcc/Warning-Options.html#index-Wimplicit-fallthrough


imanti:
Changes done. I used the marker `/* FALLTHROUGH */` that matches -Wimplicit-fallthrough levels 2, 3 and 4.

PR #12, opened 2018-09-13
Persistent willing

imanti:
There are two commits in this merge:

8bdc723 - 8021qaz: Honor willing value from config file
Read the willing value of pfc and ets variables directly from the config file before calling set_ieee_hw. In case such value exists and its value is false, then we don't set any pfc/ets data to write to hw.

9858409 - 8021qaz: Set willing to false by default 
If the willing parameter for PFC/ETSCFG is not present in lldpad.config file, then set it to false by default.

### DIFF
--- a/lldp_8021qaz.c
+++ b/lldp_8021qaz.c
@@ -223,7 +223,7 @@ static int read_cfg_file(char *ifname, struct lldp_agent *agent,
 	const char *arg = NULL;
 	char arg_path[256];
 	int res = 0, i;
-	int willing, pfc_mask, delay;
+	int pfc_mask, delay;
 
 	if (agent->type != NEAREST_BRIDGE)
 		return 0;
--- a/lldp_8021qaz.c
+++ b/lldp_8021qaz.c
@@ -213,7 +213,8 @@ bool read_cfg_file_willing(char *ifname, struct lldp_agent *agent, int tlv_type)
 	res = get_config_setting(ifname, agent->type, arg_path, &willing, 
 		CONFIG_TYPE_INT);
  	
-	return (res == 0 ? willing == 1 : true); 
+	/* willing is disabled by default */
+	return (res == 0 ? willing == 1 : false); 
 }
 
 static int read_cfg_file(char *ifname, struct lldp_agent *agent,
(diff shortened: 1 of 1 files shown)


praeluceo:
Thank you for your PR, we have it under validation now.

imanti:
Hi @praeluceo, have you had chance to validate this changes?

praeluceo:
Hi @imanti yes, we have put the PR through validation, we have commentary on both commits:
8bdc723 - Validation believes there is a typo in this commit, "Compilation error:

{{lldp_8021qaz.c: In function ‘read_cfg_file’: }}
{{lldp_8021qaz.c:226:6: error: unused variable ‘willing’ [-Werror=unused-variable] }}
{{ int willing, pfc_mask, delay;}}

In commit 8bdc723{{}} the need for that variable was removed, but the declaration was not deleted."

For commit 9858409 we would like further explanation for why setting willing to false by default would be considered beneficial. For our purposes, we disagree and find that having an adapter default to the willing state is the ideal initial scenario, and when a user sets a local DCB configuration, it is at that time where it makes the most sense to also set the PF to be non-willing. A default non-willing configuration interferes with the use case where the system administrator configures a switch, and expects that all PFs attached to it will automatically follow the switch config by default, which we feel is the expected bulk of user usage.

sowminiv:
Hi some data about the comment from Ronald: 

we have customers that want to use LLDP for topology discovery but do not want to have DCBX picked up from the switch for various reasons, e.g., some switch vendors have bugs where the advertise incorrect values, customers generally prefer static configuration because there is concern that the lldpad implementation has bugs, at least one prominent NIC vendor actually print scary warnings like "****** WARNING: lldpad service is running and may overwrite your settings ******".  Also, one can clobber the PFC config on the port by just spamming it with a spoofed lldp adv.

For such users, they end up needed to (a) start up lldpad, (b) set willing to false. There is a time interval between (a) and (b) where PFC can be clobbered. (Or not even using lldpad for topology
discovery, where it is actually quite safe and useful)

In general, setting "willing=yes" is something that can cause some non-trivial changes to the NIC configuration and to network performance in general, and it is better if this is done  very deliberately, at the discretion of the administrator. 

Hope that helps.

imanti:
@praeluceo, I pushed a new commit to delete unused 'willing' variable from first commit.

dmertman:
Can you please separate out commit [9858409](https://github.com/intel/openlldp/commit/98584099f4e30b5a7b237f227128b1c8d2bdce03) to a separate pull request, as I would like to commit the other two (get willing state from config and the fix for the unused variable in the header).

I would like further dialogue on the default state of the willing bit.

imanti:
Closing this merge request as is no longer applicable.
the commit 8bdc7236bf10f0ca389fffb56691f2a2ca1c712f is already present in master.

PR #13, opened 2018-10-08
Persistent willing2

imanti:
This merge contains only changes to honor willing value from config file; The default willing value (which is enabled) has not been changed. 

There are two commits in this merge:

189f63a - 8021qaz: Honor willing value from config file
Read the willing value of pfc and ets variables directly from the config file before calling set_ieee_hw. In case such value exists and its value is false, then we don't set any pfc/ets data to write to hw.

59c9a54 - 8021qaz: Remove unused variable
Previous commit had introduced an unused variable.

### DIFF
--- a/lldp_8021qaz.c
+++ b/lldp_8021qaz.c
@@ -203,6 +203,19 @@ static void set_ets_tsa_map(const char *arg, u8 *tsa_map)
 	free(argcpy);
 }
 
+bool read_cfg_file_willing(char *ifname, struct lldp_agent *agent, int tlv_type)
+{
+	char arg_path[256];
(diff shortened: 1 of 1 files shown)


imanti:
@dmertman 

As requested in PR #12, this PR does not modify the default willing value.

dmertman:
Ron,
Could you do a quick validation on this for me?
Thanks

imanti:
@dmertman Any news with the validation?

praeluceo:
Validation has confirmed that this PR builds fine and is functionally the same as what we were testing previously. Looks good.

PR #14, opened 2018-11-15
rpm-spec: Add vdp22 man files to lldpad.spec.in

imanti:
Add missing files man3/liblldp_clif-vdp22.* to file lldpad.spec.in
at '%files' section.


praeluceo:
FYI this PR is in our validation lab, however I do not expect any progress until after the Thanksgiving holiday.

PR #15, opened 2018-11-15
autoconf: Add systemd support in configure.ac

imanti:
Changes to let rpm use systemd or sysV depending on the target system
configuration.

If systemd is installed in the build system then variables
SPEC_BUILD_REQUIRES_[POST|PREUN|POSTUN] are filled with the string
'systemd', and variables SPEC_FILE_LLDPAD_[SERVICE|SOCKET] point to
files lldpad.service and lldpad.socket.

If systemd is not present in the build system then variables
SPEC_BUILD_REQUIRES_[POST|PREUN|POSTUN] are filled with 'chkconfig'
commands, and variables SPEC_FILE_LLDPAD_[SERVICE|SOCKET] point to
file '%{_sysconfdir}/init.d/lldpad'.

In both cases, file lldpad.spec.in have references to above variables.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -25,6 +25,39 @@ then
  AC_MSG_ERROR([no suitable flex found. Please install the 'flex' package.])
 fi
 
+dnl Begin determine the systemd use and location
+PKG_CHECK_MODULES([SYSTEMD], [systemd], use_systemd=yes, use_systemd=no)
+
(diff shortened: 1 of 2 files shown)


praeluceo:
FYI this PR is in our validation lab, however I do not expect any progress until after the Thanksgiving holiday.

imanti:
I tested it in OL7 (UEK5) and ubutnu 18.04.

PR #16, opened 2018-11-15
rpm: Add fields to *.service and *.socket files

imanti:
Add missing rules in file lldpad.service:
- Requires=lldpad.socket   # Bring up socket when service goes up
- Restart=always           # Attempt to start service when is down

and missing information in file lldpad.socket:
- Add [Unit] section with description of socket

### DIFF
--- a/lldpad.service
+++ b/lldpad.service
@@ -1,9 +1,11 @@
 [Unit]
 Description=Link Layer Discovery Protocol Agent Daemon.
 After=syslog.target network.target
+Requires=lldpad.socket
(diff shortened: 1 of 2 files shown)


praeluceo:
FYI this PR is in our validation lab, however I do not expect any progress until after the Thanksgiving holiday. That said, I don't see any reason why we can't just merge this PR, so I'm pinging @dmertman to see if we can just pull this one in, it's been a frustrating nuance of the service since the beginning.

PR #17, opened 2018-11-15
8021qaz: Print dscp2prio map

imanti:
Following definition of DSCP APP TLV with selector 5, print dscp2prio
map for DSCP APP entries instead of printing them as any other APP
entry.

DSCP standard 802.1Qcd-2015/D.2.15.3


praeluceo:
FYI this PR is in our validation lab, however I do not expect any progress until after the Thanksgiving holiday. This is a mildly larger lift to test, we do have a DSCP enabled switch in the lab, but configuration will take a little more effort than with the other PRs.

PR #19, opened 2018-12-04
lldp: Allow lldptool to modify optional TLV's content

imanti:
Add functions [get, set, test]_arg_info in lldp_basman_cmds.c to
allow users to get or set the value of optional lldp TLVs.

Optional tlvs are (lldptool tlv name inside brackets):

[portDesc] Port description
[sysName] System name
[sysDesc] System description
[sysCap] System capabilities
[mngAddr] Management address

The format to modify a TLV content is:
	lldptool -i <ifname> -T -V sysDesc info="New Value"

### DIFF
--- a/include/lldp_basman_clif.h
+++ b/include/lldp_basman_clif.h
@@ -31,7 +31,7 @@ struct lldp_module *basman_cli_register(void);
 void basman_cli_unregister(struct lldp_module *);
 int basman_print_tlv(u32, u16, char *);
 
-#define ARG_IPV4_ADDR "ipv4"
-#define ARG_IPV6_ADDR "ipv6"
-
+#define ARG_IPV4_ADDR	"ipv4"
+#define ARG_IPV6_ADDR	"ipv6"
+#define ARG_TLVINFO	"info"
 #endif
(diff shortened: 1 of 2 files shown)


praeluceo:
From Validation:
There is a compilation error after applying this patch on top of master:

```
mv -f $depbase.Tpo $depbase.Po
lldp_basman_cmds.c: In function ‘_set_arg_info’:
lldp_basman_cmds.c:238:19: error: implicit declaration of function ‘malloc’ [-Werror=implicit-function-declaration]
{{ value = (char *) malloc(sizeof(char) * size);}}
{{ ^~~~~~}}
lldp_basman_cmds.c:238:19: error: incompatible implicit declaration of built-in function ‘malloc’ [-Werror]
lldp_basman_cmds.c:238:19: note: include ‘<stdlib.h>’ or provide a declaration of ‘malloc’
cc1: all warnings being treated as errors
make: *** [Makefile:1099: lldp_basman_cmds.o] Error 1
``` 

Looks like stdlib wasn't included, validation requests if could you fix the above issue, and maybe look at that malloc and not issuing free against it?

imanti:
Ronald, sorry for the delay!

The validation that I do with the info string is unnecessary because libconfig will store it as a string (enclosed by quotes) so there is no need to delete any character. This would mean that the extra variable and mallco() are not needed.

If it works for you, let me close this PR and create a new branch/PR without this validation.

praeluceo:
Closing PR in favor of PR#20

PR #18, opened 2018-12-04
8021qaz: Add parameter passive_8021qaz to run in read only mode

imanti:
Add parameter passive_8021qaz to change 8021qaz module behavior
with respect to hw values. If passive_8021qaz is true, then hw
values will be updated/modified. If passive_dcbx is false, then
there will be no changes to hw.

To run in passive mode use -P option: 'lldpad -P'

PR #22, opened 2019-01-31
8021qaz: Add parameter read_only_8021qaz to run in read only mode

imanti:
Add parameter read_only_8021qaz to change 8021qaz module's behavior
with respect to hw values.

When read_only_8021qaz is false, then hw values get updated/modified
following the regular rules of the protocol.

If read_only_dcbx is true, then there will be no changes to hw and
user won't be allowed to call any 8021qaz 'set' functions. If lldptool
sends a 'set' request then the answer will be 'cmd_no_access'.

To run lldpad in read-only  mode use -R option: 'lldpad -R'

### DIFF
--- a/lldp_8021qaz.c
+++ b/lldp_8021qaz.c
@@ -50,6 +50,7 @@
 
 struct lldp_head lldp_head;
 struct config_t lldpad_cfg;
+extern bool read_only_8021qaz;
 
 static int ieee8021qaz_check_pending(struct port *port, struct lldp_agent *);
 static void run_all_sm(struct port *port, struct lldp_agent *agent);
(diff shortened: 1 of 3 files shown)


imanti:
This PR is a continuation of #18 after changing name of variable 'passive_8021qaz' to 'read_only_8021qaz'.

@dmertman, @paravmellanox, I decided to create this new PR with fixes for your reviews, hope that is ok for your reviews.

paravmellanox:
@imanti why did you open the new PR.
Can you please update the old PR for the code. You can force push the code that address review comments so same PR will get updated with new code. This way everything is tracked to same PR.

imanti:
> 
> 
> @imanti why did you open the new PR.
> Can you please update the old PR for the code. You can force push the code that address review comments so same PR will get updated with new code. This way everything is tracked to same PR.

I was not sure about force pushing thing! I'll do that then.

imanti:
Closing this PR. Force pushing review changes to branch in PR #18

Issue #25, opened 2019-03-20
New string compilation warnings seen with GCC 9


marxin:
This is follow up of #23, where I still see quite some warnings:
```
In file included from /usr/include/string.h:494,
                 from event_iface.c:36:
In function ‘strncpy’,
    inlined from ‘event_if_decode_rta’ at event_iface.c:80:3,
    inlined from ‘event_if_decode_nlmsg’ at event_iface.c:260:4,
    inlined from ‘event_if_process_recvmsg’ at event_iface.c:334:2,
    inlined from ‘event_iface_receive’ at event_iface.c:412:2:
/usr/include/bits/string_fortified.h:106:10: warning: ‘__builtin_strncpy’ specified bound 16 equals destination size [-Wstringop-truncation]
  106 |   return __builtin___strncpy_chk (__dest, __src, __len, __bos (__dest));
      |          ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
In file included from /usr/include/string.h:494,
                 from lldp/l2_packet_linux.c:29:
In function ‘strncpy’,
    inlined from ‘l2_packet_init’ at lldp/l2_packet_linux.c:172:2:
/usr/include/bits/string_fortified.h:106:10: warning: ‘__builtin_strncpy’ specified bound 17 equals destination size [-Wstringop-truncation]
  106 |   return __builtin___strncpy_chk (__dest, __src, __len, __bos (__dest));
      |          ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
In function ‘strncpy’,
    inlined from ‘l2_packet_init’ at lldp/l2_packet_linux.c:186:2:
/usr/include/bits/string_fortified.h:106:10: warning: ‘__builtin_strncpy’ output may be truncated copying 16 bytes from a string of length 16 [-Wstringop-truncation]
  106 |   return __builtin___strncpy_chk (__dest, __src, __len, __bos (__dest));
      |          ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
lldp_mand.c: In function ‘mand_bld_ip_chassis’:
lldp_mand.c:115:30: warning: taking address of packed member of ‘union <anonymous>’ may result in an unaligned pointer value [-Waddress-of-packed-member]
[...]

PR #27, opened 2019-04-25
Come up with STRNCPY_TERMINATED (#25).

marxin:
And use it in order to deal with new GCC 9 warnings.

### DIFF
--- a/dcb_protocol.c
+++ b/dcb_protocol.c
@@ -75,7 +75,7 @@ void pg_insert(struct pghead *head, char *ifname, pg_attribs *store)
 	entry = (struct pg_store1 *)malloc(sizeof(struct pg_store1));
 	if (!entry)
 		return;
-	strncpy(entry->ifname, ifname, sizeof(entry->ifname));
+	STRNCPY_TERMINATED(entry->ifname, ifname, sizeof(entry->ifname));
 	entry->second = store;
 	LIST_INSERT_HEAD(head, entry, entries);
 }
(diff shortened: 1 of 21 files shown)


praeluceo:
Thank you for the PR, we are validating it now.

dcb_protocol.c
```diff
@@ -2274,9 +2274,7 @@ cmd_status get_bwg_descrpt(char *device_name, u8 bwgid, char **name)
 				sizeof(char);
 			*name = (char*)malloc(size);
 			if (*name != NULL) {
-				strncpy(*name,
-					attribs.descript.pgid_desc[bwgid],
-					size); /* Localization OK */
+				memcpy(*name, attribs.descript.pgid_desc[bwgid], size); /* Localization OK */
```
dmertman: Why is this instance changed to a memcpy when all of the other ones were changed to use the new MACRO?

marxin: Doing that I would get:
```
In function ‘strncpy’,
    inlined from ‘get_bwg_descrpt’ at dcb_protocol.c:2277:5:
/usr/include/bits/string_fortified.h:106:10: error: ‘__builtin_strncpy’ output truncated before terminating nul copying as many bytes from a string as its length [-Werror=stringop-truncation]
  106 |   return __builtin___strncpy_chk (__dest, __src, __len, __bos (__dest));
      |          ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
dcb_protocol.c: In function ‘get_bwg_descrpt’:
dcb_protocol.c:2272:16: note: length computed here
 2272 |    size = (int)strlen(
      |                ^~~~~~~
 2273 |     attribs.descript.pgid_desc[bwgid]) +
      |     ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
cc1: all warnings being treated as errors
```

marxin:
Just a note that GCC `9.1.0` has been just released.

praeluceo:
This patch cleared internal validation.

Issue #11, opened 2018-09-05
New release anytime soon?


srivatsabhat:
It looks like the last release (v1.0.1) was back in 2015. Is there a plan to release a new version soon, incorporating all the improvements since then? If there isn't, please consider this as a request for one :-)

Thank you!

praeluceo:
Yes, we have a plan to bump the version string and update the documentation. We'll probably end up bumping the version before updating the documentation however. I will accept this as a request for a version change and work on getting that done so we have a place to compare all of these recent changes against.

ffontaine:
It would be really nice to get a new release. Do you still plan to release any time soon?

orgcandman:
A new release branch is available.  After a period of validation, we can officially announce it

PR #7, opened 2018-06-21
OID Printing changes

orgcandman:
The OID display from lldp-tool is currently dumping the raw bytes from the MIB, rather
than formatting them into a common dotted notation.  This series corrects that by displaying
as the common notation.

Additionally, this series incorporates the GNU Autotools 'make check' facility.  It introduces a check
in 2/3 which fails, and then fixes the failure in 3/3.

Signed-off-by: Aaron Conole <aconole@redhat.com>

Issue #1, opened 2017-12-11
open-lldp migration


praeluceo:
Migration tasks:

- [x] Migrate http://open-lldp.org/git/ to Github
- [x] Work on migrating existing mailing list patches to Github
- [x] Ensure that there are no gaps between this new repository and https://github.com/jrfastab/lldpad
- [ ] Notify OSVs of migration and new canonical source of open-lldp source

praeluceo:
Import complete

PR #30, opened 2019-07-31
Two bug fixes

orgcandman:
First is a fix to the lldp_util routine when pulling address information.

Second is a fix to the ecp22 module under error conditions.

Neither has been tested thoroughly.

### DIFF
--- a/lldp_util.c
+++ b/lldp_util.c
@@ -1055,9 +1055,11 @@ int get_saddr6(const char *ifname, struct sockaddr_in6 *saddr)
 
 	rc = getifaddrs(&ifaddr);
 	if (rc == 0) {
+		rc = -1;
 		for (ifa = ifaddr; ifa != NULL; ifa = ifa->ifa_next) {
-			if ((ifa->ifa_addr->sa_family == AF_INET6) &&
-			    (strncmp(ifa->ifa_name, ifname, IFNAMSIZ) == 0)) {
+			if (strncmp(ifa->ifa_name, ifname, IFNAMSIZ))
+				continue;
+			if (ifa->ifa_addr && (ifa->ifa_addr->sa_family == AF_INET6)) {
 				memcpy(saddr, ifa->ifa_addr, sizeof(*saddr));
 				rc = 0;
 				break;
--- a/qbg/ecp22.c
+++ b/qbg/ecp22.c
@@ -782,6 +782,10 @@ void ecp22_start(char *ifname)
 	ecp = find_ecpdata(ifname, eud);
 	if (!ecp)
 		ecp = ecp22_create(ifname, eud);
+	if (!ecp) {
+		LLDPAD_DBG("%s:%s failed creating ECP22 instance\n", __func__, ifname);
+		return;
+	}
 	ecp->max_retries = ECP22_MAX_RETRIES_DEFAULT;
 	ecp->max_rte = ECP22_ACK_TIMER_DEFAULT;
 	LIST_INIT(&ecp->inuse.head);


orgcandman:
@praeluceo any comments?

praeluceo:
Good morning Aaron, apologies for the delay in responding. We have this in validation, however we may need to only perform a regression test on it. Do you have testing hints, or a tracking Bugzilla we can use in validation against this patch?

orgcandman:
Yes - for the ifaddr:


Program terminated with signal 11, Segmentation fault.
#0  0x0000562b97f6dbcc in get_saddr6 (ifname=ifname@entry=0x562b993d97c0 "vethfd5c38f1", saddr=saddr@entry=0x7ffed42e4870) at lldp_util.c:1043
1043                            if ((ifa->ifa_addr->sa_family == AF_INET6) &&
(gdb) bt
#0  0x0000562b97f6dbcc in get_saddr6 (ifname=ifname@entry=0x562b993d97c0 "vethfd5c38f1", saddr=saddr@entry=0x7ffed42e4870) at lldp_util.c:1043
#1  0x0000562b97f6dc71 in get_ipaddr6 (ifname=ifname@entry=0x562b993d97c0 "vethfd5c38f1", in6=in6@entry=0x7ffed42e48e2) at lldp_util.c:1060
#2  0x0000562b97f6e863 in mand_bld_portid_tlv (md=md@entry=0x562b993d97c0, agent=0x562b993d8a60) at lldp_mand.c:382
#3  0x0000562b97f6ec85 in mand_bld_tlv (md=md@entry=0x562b993d97c0, agent=agent@entry=0x562b993d8a60) at lldp_mand.c:475
#4  0x0000562b97f6f17c in mand_ifup (ifname=0x7ffed42e4cf0 "vethfd5c38f1", agent=0x562b993d8a60) at lldp_mand.c:606
#5  0x0000562b97f555a5 in oper_add_device (device_name=device_name@entry=0x7ffed42e4cf0 "vethfd5c38f1") at event_iface.c:218
#6  0x0000562b97f55f0e in event_if_decode_nlmsg (len=<optimized out>, data=0x7ffed42e4d10, route_type=16) at event_iface.c:313
#7  event_if_process_recvmsg (nlmsg=0x7ffed42e4d00) at event_iface.c:334
#8  event_iface_receive (sock=<optimized out>, eloop_ctx=<optimized out>, sock_ctx=<optimized out>) at event_iface.c:412
#9  0x0000562b97f56504 in eloop_sock_table_dispatch (table=table@entry=0x562b981acbf0 <eloop+16>, fds=fds@entry=0x562b993d3140) at eloop.c:209
#10 0x0000562b97f56c73 in eloop_run () at eloop.c:526
#11 0x0000562b97f512ea in main (argc=<optimized out>, argv=<optimized out>) at lldpad.c:459
  

(gdb) p ifa
$1 = (struct ifaddrs *) 0x55a64c323f18
(gdb) p ifa->ifa_addr
$2 = (struct sockaddr *) 0x0



For the ecp22 fix (this is from some automated tool):


1 | ecp22_start | /usr/sbin/lldpad | 0x3cc3a       Build id: 6965d7a0bd89816fd46ad52524763484c03576ff | -
-- | -- | -- | -- | --
2 | evb22_timer | /usr/sbin/lldpad | 0x39a0a       Build id: 6965d7a0bd89816fd46ad52524763484c03576ff | -
3 | timer | /usr/sbin/lldpad | 0x17eea       Build id: 6965d7a0bd89816fd46ad52524763484c03576ff | -
4 | eloop_run | /usr/sbin/lldpad | 0xbc2f       Build id: 6965d7a0bd89816fd46ad52524763484c03576ff | -
5 | main | /usr/sbin/lldpad | 0x62ca       Build id: 6965d7a0bd89816fd46ad52524763484c03576ff

lldp_util.c
```diff
@@ -1055,9 +1055,11 @@ int get_saddr6(const char *ifname, struct sockaddr_in6 *saddr)
 
 	rc = getifaddrs(&ifaddr);
 	if (rc == 0) {
+		rc = -1;
 		for (ifa = ifaddr; ifa != NULL; ifa = ifa->ifa_next) {
-			if ((ifa->ifa_addr->sa_family == AF_INET6) &&
-			    (strncmp(ifa->ifa_name, ifname, IFNAMSIZ) == 0)) {
```
sregister: The open bracket is deleted but the corresponding closing bracket isn't touched. Should the open bracket be added to one of the other lines?

lldp_util.c
```diff
@@ -1055,9 +1055,11 @@ int get_saddr6(const char *ifname, struct sockaddr_in6 *saddr)
...
+			if (strncmp(ifa->ifa_name, ifname, IFNAMSIZ))
+				continue;
+			if (ifa->ifa_addr && (ifa->ifa_addr->sa_family == AF_INET6))
 				memcpy(saddr, ifa->ifa_addr, sizeof(*saddr));
 				rc = 0;
```
sregister: These lines are indented as if they are part of the if block but I don't think that was the intention. This is throwing misleading-indentation errors on gcc 9.1.

lldp_util.c
penguin359: A strncmp() should always compared against zero as it returns a tristate value and not a boolean. It is much harder to determine the author's intent with that comparison is omitted.

lldp_util.c
orgcandman: Correct - I missed it.  Will address in v2

lldp_util.c
orgcandman: In this case we only care that the interface is exactly the same name, so while strncmp will tell us which side of the string was 'greater' it isn't relevant for the code in question.  Indeed, a common idiom is:
  if (!strncmp(...))
rather than an explicit test for 0, since the 'not' is readable, conserves horizontal space, and conveys the same information.  If it mattered which side of the string were greater (for example because the code were making decisions based on a tree structure) then an explicit check for >, <, and = would make sense.  Here it does not.

orgcandman: Yes.  I'll fix it as I address the other comment.

orgcandman:
Updated.

praeluceo:
Thanks! We'll get it reevaluated.

lldp_util.c
```diff
@@ -1055,9 +1055,11 @@ int get_saddr6(const char *ifname, struct sockaddr_in6 *saddr)
 
 	rc = getifaddrs(&ifaddr);
 	if (rc == 0) {
+		rc = -1;
 		for (ifa = ifaddr; ifa != NULL; ifa = ifa->ifa_next) {
-			if ((ifa->ifa_addr->sa_family == AF_INET6) &&
-			    (strncmp(ifa->ifa_name, ifname, IFNAMSIZ) == 0)) {
+			if (strncmp(ifa->ifa_name, ifname, IFNAMSIZ))
```
penguin359: I would still recommend using an explicit comparison to zero on the strncmp() for code readability. If the goal is to state not matching, then != 0 avoids any ambiguity.

orgcandman:
Anything else preventing merge?  Customers are hitting these errors, so I'd like to get it merged so we can deliver fixes (rather than rely on custom hotfix builds).

PR #32, opened 2019-09-05
Misc static analysis fixes

orgcandman:
These are some static analysis and compiler warning fixes.

Tested with clang and gcc.  There are additional issues around enum conversion, but I'll submit them separately for evaluation.

### DIFF
--- a/lldp_dcbx_nl.c
+++ b/lldp_dcbx_nl.c
@@ -801,17 +801,6 @@ int set_hw_app(char *ifname, appgroup_attribs *app_data)
 	return(recv_msg(DCB_CMD_SAPP, DCB_ATTR_APP, seq));
 }
 
-int run_cmd(char *cmd, ...)
-{
-	char cbuf[128];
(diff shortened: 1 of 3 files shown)


lldp_dcbx_nl.c
```diff
@@ -801,17 +801,6 @@ int set_hw_app(char *ifname, appgroup_attribs *app_data)
 	return(recv_msg(DCB_CMD_SAPP, DCB_ATTR_APP, seq));
 }
 
-int run_cmd(char *cmd, ...)
-{
-	char cbuf[128];
-	va_list args;
-
-	va_start(args, cmd);
-	vsprintf(cbuf, cmd, args);
```
penguin359: Even if this function was used, it could use some improvement. Writing to a char array with vsprintf() provides no bounds checking and should definitely be replace. Good to just remove it all.

orgcandman:
@dmertman ping

PR #33, opened 2019-09-16
Travis CI Integration

penguin359:
This branch adds in support to run a build and test on OpenLLDP automatically through Travis CI. If enabled on the main repo, this should help ensure pull requests actually compile and pass some level of testing first.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,7 +2,7 @@ language: c
 before_install: sudo apt-get update
 install: sudo apt-get install -y libconfig-dev libnl-3-dev rpm
 before_script: ./bootstrap.sh
-script: ./configure && make && make test && sudo make install && ./contrib/build-rpm.sh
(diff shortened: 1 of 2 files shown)


penguin359:
I've updated the build Matrix to use GCC 4.8 and also added GCC 8 and 9. GCC 4.8 corresponds to what is provided with RHEL 7.x and GCC 8.2 is included with RHEL 8.x. This should provide a little better coverage for those releases.

orgcandman:
I have a different version at:

https://github.com/orgcandman/openlldp/tree/detect_another

It requires a number of fixes to support clang (I've submitted some at https://github.com/intel/openlldp/pull/32)

I prefer most of these travis changes to be squashed into a single commit.  Do you think it makes sense to split them up - doesn't seem like it adds any value.

PR #35, opened 2019-10-10
Preserve physical mac address for bonding slaves

hreinecke:
Bonding slaves should transmit their permanent mac address in the Port ID TLV to allow the different ports to be distinguished on the switch side.


praeluceo:
Thank you for the PR, we have it under review now!

PR #38, opened 2019-11-25
dcbx: check for dcbx_data result

orgcandman:
The dcbx_data function can return a null pointer, but the dcbx tlv function
will happily attempt to dereference.  Prevent this by checking for result
and bailing.

Fixes: a37b7e0f3b66 ("lldpad: initial git commit")


orgcandman:
ping?

PR #40, opened 2020-02-11
Revert two

dmertman:
Backing out two patches that broke openlldp functionality


aron-silverton:
Is not possible fix commit 189f63a0b72b rather than revert it?

Nits:
I don't think the PR should include a merge commit.

You may want to consider rewording this sentence in the commit message

> Two patches were added to try and honor the
willing value got ets and pfc from the config
file.

And referencing the commits you are reverting with additional context, for example:
```
59c9a549831a ("8021qaz: Remove unused variable")
189f63a0b72b ("8021qaz: Honor willing value from config file")
```

Issue #34, opened 2019-10-08
Unable to apply DCB configuration


sregister:
When using tip of tree 3168e11 I'm unable to apply a dcb map by running:
```
lldptool -Ti $1 -V ETS-CFG enableTx=yes willing=no up2tc=0:0,1:0,2:0,3:1,4:2,5:0,6:0,7:0 tcbw=10,30,60 tsa=0:ets,1:ets,2:ets
lldptool -Ti $1 -V ETS-REC enableTx=yes up2tc=0:0,1:0,2:0,3:1,4:2,5:0,6:0,7:0 tcbw=10,30,60 tsa=0:ets,1:ets,2:ets
lldptool -Ti $1 -V PFC enable=yes enableTx=yes willing=no enabled=3,4
```

The above commands return success:
```
enabled = yes
willing = no
up2tc = 0:0,1:0,2:0,3:1,4:2,5:0,6:0,7:0
tcbw = 10% 30% 60% 0% 0% 0% 0% 0% 
TSA = 0:ets 1:ets 2:ets 3:strict 4:strict 5:strict 6:strict 7:strict 
enabled = yes
up2tc = 0:0,1:0,2:0,3:1,4:2,5:0,6:0,7:0
tcbw = 10% 30% 60% 0% 0% 0% 0% 0% 
TSA = 0:ets 1:ets 2:ets 3:strict 4:strict 5:strict 6:strict 7:strict 
enabled = yes
willing = no
prio = 3,4
```

But the DCB map never applies. When monitoring the driver the dcbnl set functions never seem to be hit. After bisecting it was determined that commits 59c9a54 189f63a seem to cause this regression.

orgcandman:
Sorry, I just saw this.  Can you try the following patch?

```
diff --git a/lldp_8021qaz.c b/lldp_8021qaz.c
index 219db80..09e9060 100644
--- a/lldp_8021qaz.c
+++ b/lldp_8021qaz.c
@@ -214,7 +214,7 @@ bool read_cfg_file_willing(char *ifname, struct lldp_agent *agent, int tlv_type)
 	res = get_config_setting(ifname, agent->type, arg_path, &willing, 
 		CONFIG_TYPE_INT);
  	
-	return (res == 0 ? willing == 1 : true); 
+	return (res == 0 ? !!willing : true);
 }
 
 static int read_cfg_file(char *ifname, struct lldp_agent *agent,
```

[...]

PR #41, opened 2020-02-11
8021qaz: Revert patches attemping to honor config file

dmertman:
Two patches were added to try and honor the
willing value for ets and pfc from the config
file.  This change caused lldpad to stop
accepting new configurations from the commnad
line tools after the first configuration.

Backing out these two patches to restore
functionality of lldpad.

Hashes of patches being reverted:
59c9a549831ad9bd3baac0d88dce2f724e045113
"8021qaz: Remove unused variable"

189f63a0b72bdd70710bfcce2a104e0ad7e915c1
"8021qaz: Honor willing value from config file"

Signed-off-by: Dave Ertman <david.m.ertman@intel.com>
Change-type: Other
Title: Revert patches attemping to honor config file

### DIFF
--- a/lldp_8021qaz.c
+++ b/lldp_8021qaz.c
@@ -204,37 +204,36 @@ static void set_ets_tsa_map(const char *arg, u8 *tsa_map)
 	free(argcpy);
 }
 
-bool read_cfg_file_willing(char *ifname, struct lldp_agent *agent, int tlv_type)
-{
-	char arg_path[256];
-	int res, willing = 0;
-
-	snprintf(arg_path, sizeof(arg_path), "%s%08x.%s", TLVID_PREFIX, 
-		TLVID_8021(tlv_type), ARG_WILLING);
-	res = get_config_setting(ifname, agent->type, arg_path, &willing, 
-		CONFIG_TYPE_INT);
- 	
-	return (res == 0 ? willing == 1 : true); 
-}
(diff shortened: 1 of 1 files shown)


sregister:
I think the objective of 189f63a was to work around an issue where openlldp was changing dcb parameters after it was set using a different tool: http://lists.open-lldp.org/pipermail/lldp-devel/2018-December/001306.html. The problem is the patch breaks applying non-willing dcb configs; see issue https://github.com/intel/openlldp/issues/34

If the user wants to use an application like mlnx_qos to apply dcb parameters but retain LLDP for link layer discovery one option would be to use https://github.com/vincentbernat/lldpd which provides LLDP functionality but does not do DCBx. Alternatively if 189f63a was to be revisited, I think the approach should be implement a new optional flag that would _totally_ disable the IEEE/CEE state machines so openlldp never hits the dcbnl api.

aron-silverton:
Thanks, @sregister. I did not follow the original discussion around those patches when they were submitted.

I think the better solution is if vendors do not require proprietary tools to configure DCB. IMHO, configuring NICs should be done with upstream tools whether openlldp or functionality currently in or added in the future to iproute2.

sregister:
I can confirm issue https://github.com/intel/openlldp/issues/34 is fixed when using this branch.

PR #57, opened 2020-09-10
lldp: do not call 'assert' in rxProcessFrame()

gonzoleeman:
The state machine might have called 'DELETE' just before RX got
triggered, so we might end up having no frame to process.
If so just ignore this packet.

Signed-off-by: Hannes Reinecke <hare@suse.de>

### DIFF
--- a/lldp/rx.c
+++ b/lldp/rx.c
@@ -139,7 +139,16 @@ void rxProcessFrame(struct port *port, struct lldp_agent *agent)
 	int err;
 	struct lldp_module *np;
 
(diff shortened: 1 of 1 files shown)


gonzoleeman:
This is the first of a few pull requests I'm going to submit from the openSUSE downstream repository here on github (openSUSE/lldpad)

penguin359 (approved):
This looks far safer to do than a stray assert lingering in the code.

orgcandman:
Thanks, Hannes & Loren.

Merged.

PR #61, opened 2020-09-23
travis-ci: Add gcc10 check

sregister:
Note gcc-10 checks are being done on Bionic instead of Focal since there is a
bug in Focal's version of gcc-10 (at time of writing) which was reporting false 
failures: https://gcc.gnu.org/bugzilla/show_bug.cgi?id=94647

Signed-off-by: Scott Register <scottx.register@intel.com>

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -47,3 +47,12 @@ matrix:
             - gcc-9
       env:
         - CC="gcc-9"
+    - addons:
+        apt:
+          sources:
+            - sourceline: 'ppa:ubuntu-toolchain-r/test'
+          packages:
+            - gcc-10
+      env:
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -1,6 +1,5 @@
-language: c
-dist: xenial
```
orgcandman, line 2: Do we need to actually change these?  I think the individual blocks can override the global settings.  Is there an advantage to one vs. the other?

sregister: I just checked this and you are right; I've moved dist xenial back to global scope. My branch (and this PR) have been updated.  Thanks

orgcandman:
Thanks!

PR #62, opened 2020-10-01
Build flags changes

orgcandman:


orgcandman:
@ffontaine and @kloczek - this isn't exactly what we've discussed, but I feel it's the best solution. Developers still get warnings on by default, and distros can enable warnings later.  I agree with the original approach by @ffontaine after looking over hundreds of projects doing the same thing.

orgcandman:
@sregister please review patch 1.

sregister (approved):
LGTM; I can see the tests are actually being executed in the CI terminal log. CC @penguin359

kloczek:
As it is not possible to remove injecting compiler warning flags in autoconf IMO this is only movinf those injections from one place to another.
Really .. instead polluting generic build process better would be move those flags to .travis.yml and/or bootstrap.sh.

As I wrote with that patch it woild be not possible to make test build with injected in $CFLAGS env variable exact dset of warning flags or just single warning flag.

orgcandman:
Would adding a disable/enable for the warnings be okay?

orgcandman:
I've changed patch 2/2 to include a flag "--{enable,disable}-warnings" which will turn off the warning flags.  If this is acceptable, I will merge.

orgcandman:
I am merging this.  If something needs changing, let me know.

PR #64, opened 2020-10-23
Lldp head fixes

orgcandman:
Fix a segfault when starting lldpad due to module cleanup.


orgcandman:
I am merging this series on Thursay, and backporting 2/3 and 3/3 to branch-1.1

PR #69, opened 2020-12-08
tx: when operating in rx-only mode don't send a port shutdown pdu

orgcandman:
Currently, lldpad correctly transmits a shutdown pdu with ttl = 0s when
transitioning from rxtx or txonly to rxonly.  However, when we shutdown
lldpad it will transmit a shutdown pdu even if the port is configured
to rxonly mode.  For some implementations of LLDP this can create
a confusing state and lead to issues in the network.  Correct this by
only transmitting a shutdown PDU when going from any transmit mode to
a receive only mode, and don't transmit PDUs on shutdown if the port
agent isn't configured to transmit.

Reported-by: Matthew Whitehead <mwhitehe@redhat.com>
Reported-at: https://bugzilla.redhat.com/show_bug.cgi?id=1905210

### DIFF
--- a/lldp/tx.c
+++ b/lldp/tx.c
@@ -270,7 +270,7 @@ void run_tx_sm(struct port *port, struct lldp_agent *agent)
 			process_tx_idle(agent);
 			break;
 		case TX_SHUTDOWN_FRAME:
-			process_tx_shutdown_frame(port, agent);
+			process_tx_shutdown_frame(port, agent, true);
 			break;
 		case TX_INFO_FRAME:
 			process_tx_info_frame(port, agent);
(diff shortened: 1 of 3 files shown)


orgcandman:
@gonzoleeman or @penguin359 can you take a look?

lldp/tx.c
```diff
@@ -330,8 +330,19 @@ void process_tx_idle(UNUSED struct lldp_agent *agent)
 	return;
 }
 
-void process_tx_shutdown_frame(struct port *port, struct lldp_agent *agent)
+/* we ignore 'state' value in the case that we have recently transitioned
+ * to the shutdown state (in the case of the 'tx' state change) to allow
+ * for transmitting the ttl==0 as required by the IEEE standard. */
+void process_tx_shutdown_frame(struct port *port, struct lldp_agent *agent,
+				bool ignoreState)
```
gonzoleeman, line 337: A nit: you call it ignoreState, but you are ignoring status. It would be nice if the sense of the two linked things were the same (speaking as one that doesn't know the code well).

gonzoleeman (approved):
I'd like to see the variable name change, but if there's no other changes, you can choose to leave it the same if you wish.

gonzoleeman:
> @gonzoleeman or @penguin359 can you take a look?

done

lldp/tx.c
orgcandman: Okay - if it's good with that nit corrected, I can fix and commit with your Acked-by if you give it.

orgcandman:
ugh... sorry I forgot to command amend and accidentally pushed this.  I will submit a patch to fix the variable name.

PR #71, opened 2021-03-08
agent: reset frame status on message delete  Currently, when the agent state machine transitions out of DELETE_INFO, it leaves the rcvFrame flag set. This flag should be cleared since the frame info is no longer considered usable.  Signed-off-by: Rajesh B M bmrajesh@gmail.com

rajeshm-elisity:
There could be a core at 
assert(agent->rx.framein && agent->rx.sizein); 
in rxProcessFrame().
This is most likely due to state transition from DELETE_INFO to RX_FRAME.
When we delete a frame in process_delete_info() we should also set:
agent->rx.rcvFrame = false;

Thanks & Regards,
Rajesh

### DIFF
--- a/lldp/rx.c
+++ b/lldp/rx.c
@@ -568,6 +568,7 @@ void process_delete_info(struct port *port, struct lldp_agent *agent)
 
 	agent->rx.sizein = 0;
 	agent->rx.remoteChange = true;
+	agent->rx.rcvFrame = false;
 	return;
 }
 


orgcandman:
Looks good - can you please change your commit message:

"agent: reset frame status on message delete

Currently, when the agent state machine transitions out of DELETE_INFO, it leaves the
rcvFrame flag set.  This flag should be cleared since the frame info is no longer considered
usable.

Signed-off-by: Rajesh B M <email@here.com>
"

And I will merge it and backport to the stable branch.

rajeshm-elisity:
Thank you, I have changed the message.

orgcandman:
Pushed to 'master' - let me know if we need this on branch-1.1 or if you think it's fine to omit.

rajeshm-elisity:
Hi,

It’s ok to omit for now as the earlier commit replaced assert in rxProcessFrame() with NULL check.
It would be good though to push it, in case there are any other places this change might affect or in future there is some code that relies on this change.

Thanks & Regards,
Rajesh



From: Aaron Conole ***@***.***>
Reply to: intel/openlldp ***@***.***>
Date: Monday, 5 April 2021 at 9:36 PM
To: intel/openlldp ***@***.***>
Cc: Rajesh M ***@***.***>, Author ***@***.***>
Subject: Re: [intel/openlldp] agent: reset frame status on message delete Currently, when the agent state machine transitions out of DELETE_INFO, it leaves the rcvFrame flag set. This flag should be cleared since the frame info is no longer considered usable. ...


Pushed to 'master' - let me know if we need this on branch-1.1 or if you think it's fine to omit.

PR #72, opened 2021-04-02
Avoiding null pointer dereference

binho1991:
This pull request aims to fix null pointer dereference.

[Null pointer dereference](https://cwe.mitre.org/data/definitions/476.html)

### DIFF
--- a/ctrl_iface.c
+++ b/ctrl_iface.c
@@ -180,6 +180,8 @@ int clif_iface_attach(struct clif_data *clifd,
 	} else {
 		tlv = strdup(ibuf);
 		str = tlv;
+		if (!str)
+			goto err_tlv;
 		str++;
 		/* Count number of TLV Modules */
 		tokenize = strtok(str, delim);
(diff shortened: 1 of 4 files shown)


rajeshm-elisity:
I caught this by code audit.


Thanks & Regards,
Rajesh



From: Aaron Conole ***@***.***>
Reply to: intel/openlldp ***@***.***>
Date: Monday, 5 April 2021 at 6:11 PM
To: intel/openlldp ***@***.***>
Cc: Subscribed ***@***.***>
Subject: Re: [intel/openlldp] Avoiding null pointer dereference (#72)


How did you catch these? Was this by code audit? Some static analysis tool?

I would like to change some of these out. It would be nice to switch to xmalloc()/xstrdup() type calls, so that
we simply know that we have valid memory whenever the functions return. Under linux, the only case where
malloc(), etc. will return NULL is when we've requested more memory than the system can handle. Otherwise,
the overcommit mechanism is always returning a valid pointer and only when it would get faulted in will
overcommit protection kick in (which might invoke OOM killer). We still need these checks (so thanks for spotting
them), but maybe there's a more comprehensive cleanup we can do here.

PR #79, opened 2021-08-25
Static analysis scan fixes

orgcandman:
After a quick run through coverity and GCC's experimental scanner, a number of
issues were flagged.  While some were false positives, on some closer looks, it seems
that some are real issues.  None seem to be security issues, but all need to be addressed.

The bulk are errors during error processing.  These should almost never occur, but we fix
them here.

### DIFF
--- a/lldp_8021qaz.c
+++ b/lldp_8021qaz.c
@@ -1563,47 +1563,62 @@ static bool unpack_ieee8021qaz_tlvs(struct port *port,
 	/* Process */
 	switch (tlv->info[OUI_SIZE]) {
 	case IEEE8021QAZ_ETSCFG_TLV:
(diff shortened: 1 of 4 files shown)


orgcandman:
@penguin359 please also have a look

orgcandman:
@hreinecke maybe you can provide a quick spot check of the patches?  @sregister also?  Reviewed-by / Acked-by ?

hreinecke (approved):
Looks correct.

hreinecke (approved):
Good point; looks correct.

hreinecke (approved):
Looks correct.

orgcandman:
@srlawton any news?

penguin359 (approved):
This all looks good and can help avoid some possible crashes.

PR #81, opened 2021-09-20
Revert "Use interface index instead of name in libconfig"

orgcandman:
This reverts commit e272d34f45bc15d52424228f824c1ad96932867e.

During production, it was found that the ifindex can be unstable.
See discussion at https://github.com/intel/openlldp/pull/3

### DIFF
--- a/config.c
+++ b/config.c
@@ -56,21 +56,6 @@
 config_t lldpad_cfg;
 
 /*
- * config_ifkey - Generates a config key
- *
- * Given an interface name this functions generates
- * a key (based on interface's index) suitable
- * to pass to libconfig.
- *
- */
-void config_ifkey(const char *name, char *ifkey) {
-	int index = if_nametoindex(name);
-	
-	if(index)
-		sprintf(ifkey, "if%d", index);
-}
-
-/*
  * init_cfg - initialze the global lldpad_cfg via config_init
  *
  * Returns true (1) for succes and false (0) for failed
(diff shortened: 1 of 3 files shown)


orgcandman:
NOTE: I haven't done any testing with this revert - this proposal needs some vetting before merging it.

orgcandman:
@penguin359 see the discussion - I think @srlawton and @dmertman agreed we need to revert this patch and work on a new proposal for handling interface lookups from configurations.  There is no guarantee about ifindex and some naming daemons (like those present in systemd) can make ifnames not match with ifindex on boot.  This popped up at a customer site.  Maybe @gonzoleeman can also look?

orgcandman:
There is another good reason to revert this change, the test cases in the test/ directory rely on ifname.

I plan to apply this revert in a week and push if I don't get any response.

penguin359:
Yes, ifindex is not designed to be a unique identifier across reboots. Instead, it's designed to be the unique ID for an interface to follow during a single boot of the kernel. During a single boot of the kernel, they are given out serially and used only once. It will track the interface from the first instance it's registered as a netdev and through any configuration and renaming such as when SystemD renamed eth0 to enp2s3f0 and so on until the netdev is unregistered or system shutdown. If a driver for an interface is unloaded and reloaded, the old interface index will not be re-used and a new one will be given to it. Also, since they are given out serially, it's possible a new NIC added between boots will be discovered in between the existing list of interfaces bumping later ones up. The only way to truly track an interface across system reboots is by relying on either it's hardware MAC address or it interface name using the SystemD Predictable Network Interface Names.

The ifindex can be useful for tracking an interface across any renaming that the user might choose to do during this boot, but it can't be relied upon for tracking the interface beyond the life of the current boot and probably not more than the life of the daemon. It should be rediscovered at start-up from the known, previous interface name. Also, if an interface disappears because the superuser is upgrading the network driver, that interface might reappear later, but with a new ifindex since they are never re-used in one boot. It's primary use would simply be to retain a netdev's configuration after a rename.

Issue #82, opened 2021-10-28
optional TLVs are empty even with "enableTx=on"


dusane:
lldpad v 1.1.0

If I use following : 
lldptool set-lldp -i $lldp_intf adminStatus=rxtx
lldptool -T -i $lldp_intf -V sysName enableTx=yes 
lldptool -T -i $lldp_intf -V portDesc enableTx=yes 
lldptool -T -i $lldp_intf -V sysDesc enableTx=yes 
lldptool -T -i $lldp_intf -V sysCap enableTx=yes
lldptool -T -i $lldp_intf -V mngAddr enableTx=yes

I don't get following optional TLVs for interface $lldp_intf
"System Name TLV, System Description TLV, System Capabilities TLV"

The only change that helps getting these optional TLVs is to using "enableTx=no" for mngAddr.
( If I set it "enableTx=no" for any other TLV its not helping ).

Shouldnt "enableTx =yes" enable the specific TLV ?

dusane:
Has anyone seen this ? Is this a known issue ? 
The workaround for this issue is to use "lldptool -T -i $lldp_intf -V mngAddr enableTx=no"

orgcandman:
What do the debug logs say?

dusane:
The patch works !

PR #84, opened 2021-11-10
basman: use return address when pulling address

orgcandman:
The managed address pulling routine will fail to reset the return
value from a previous attempt if no IPv4 and IPv6 addresses are
available.  Use the return address of the hwaddr fetch.

Resolves: https://github.com/intel/openlldp/issues/82


orgcandman:
@penguin359 please have a look

orgcandman:
Merging - there is a successful test report on the Issues page.

Issue #87, opened 2022-10-26
coredump in lldptool 


gaoxingwang:
After the lldptool enters the interactive mode, a core dump occurs when the lldptool exits directly with ctrl+d.

Run the following command to reproduce the problem:
```
# lldpad -d
# lldptool
```
![image](https://user-images.githubusercontent.com/26615073/197776308-5ff6f57d-564d-406e-8f28-8ba070e87d91.png)

gaoxingwang:
There appears to be a null pointer dereference issue here
![image](https://user-images.githubusercontent.com/26615073/197776839-756383a2-9a15-4d16-bb52-6b131525bee6.png)

![image](https://user-images.githubusercontent.com/26615073/197776890-40d7eb18-7135-4110-bc2d-86f408741c9e.png)

PR #88, opened 2022-10-26
lldptool: fix null pointer deference

gaoxingwang:
fix https://github.com/intel/openlldp/issues/87

### DIFF
--- a/weak_readline.c
+++ b/weak_readline.c
@@ -80,10 +80,14 @@ void using_history(void)
 void close_history(void)
 {
 	if (inited) {
-		dlclose(rl_handle);
-		rl_handle = NULL;
-		dlclose(hist_handle);
-		hist_handle = NULL;
+                if (rl_handle) {
+                        dlclose(rl_handle);
+                        rl_handle = NULL;
+                }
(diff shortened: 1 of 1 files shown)


gaoxingwang:
PTAL @orgcandman

orgcandman:
How did you generate this error?  I assume that if an issue occurs during weak_readline_init then the rl_handle variable is reinitialized to NULL.  Were you able to reproduce?

I think there is room for cleanup, but I'm curious if this is cleanup or if you observed a real issue here (which would point to some kind of race condition).

gaoxingwang:
My lldpad runs on linux. So, when dlclose(NULL) , Segmentation fault happens.

orgcandman:
Thanks!

penguin359 (commented):
lgtm I suspect that this bug was introduced in commit e417cd91eecafc67bc5461ee05975ee926f1ca4f and only surfaces when lldptool/dcbtool are run in interactive mode on a system without libreadline currently installed. Just adding a review that this was accepted so that it is correctly categorized by the OpenSSF scoreboard checker.

PR #90, opened 2023-01-11
Fixes for memory leaks and other issues

rosslagerwall:
Most of these were found from Valgrind output and subsequent manual inspection of the code.

After these commits, Valgrind doesn't show any memory leaks or other memory errors for my use case.

### DIFF
--- a/lldp/rx.c
+++ b/lldp/rx.c
@@ -342,29 +342,69 @@ void rxProcessFrame(struct port *port, struct lldp_agent *agent)
 			}
 		}
 		if (tlv->type == TYPE_4) { /* port description */
-			agent->lldpdu |= RCVD_LLDP_TLV_TYPE4;
-			agent->rx.manifest->portdesc = tlv;
-			tlv_stored = true;
+			if (agent->lldpdu & RCVD_LLDP_TLV_TYPE4) {
+				LLDPAD_INFO("Received multiple port description"
+					" TLVs in this LLDPDU\n");
+				frame_error++;
+				free_unpkd_tlv(tlv);
+				goto out;
+			} else {
+				agent->lldpdu |= RCVD_LLDP_TLV_TYPE4;
+				agent->rx.manifest->portdesc = tlv;
+				tlv_stored = true;
+			}
 		}
 		if (tlv->type == TYPE_5) { /* system name */
-			agent->lldpdu |= RCVD_LLDP_TLV_TYPE5;
-			agent->rx.manifest->sysname = tlv;
-			tlv_stored = true;
+			if (agent->lldpdu & RCVD_LLDP_TLV_TYPE5) {
+				LLDPAD_INFO("Received multiple system name"
+					" TLVs in this LLDPDU\n");
+				frame_error++;
+				free_unpkd_tlv(tlv);
+				goto out;
+			} else {
+				agent->lldpdu |= RCVD_LLDP_TLV_TYPE5;
+				agent->rx.manifest->sysname = tlv;
+				tlv_stored = true;
+			}
 		}
 		if (tlv->type == TYPE_6) { /* system description */
-			agent->lldpdu |= RCVD_LLDP_TLV_TYPE6;
-			agent->rx.manifest->sysdesc = tlv;
-			tlv_stored = true;
+			if (agent->lldpdu & RCVD_LLDP_TLV_TYPE6) {
+				LLDPAD_INFO("Received multiple system description"
+					" TLVs in this LLDPDU\n");
+				frame_error++;
+				free_unpkd_tlv(tlv);
+				goto out;
+			} else {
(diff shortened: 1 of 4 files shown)


orgcandman:
Thanks for these.  I'll take a look this week, but a quick glance looks okay.

rosslagerwall:
Hello, any chance this could be reviewed? Thanks

lldp/tx.c
```diff
@@ -96,9 +96,10 @@ bool mibConstrInfoLLDPDU(struct port *port, struct lldp_agent *agent)
 	fb_offset += ptlv->size;
 	free_pkd_tlv(ptlv);
 
-	if (datasize < ETH_MIN_DATA_LEN)
+	if (datasize < ETH_MIN_DATA_LEN) {
+		memset(agent->tx.frameout + fb_offset, 0, ETH_ZLEN - fb_offset);
```
orgcandman, line 100: This fix looks good.  I guess we have the same issue in mibConstrShutdownLLDPDU as well.  Do you agree?

lldp/rx.c
```diff
@@ -342,29 +342,68 @@ void rxProcessFrame(struct port *port, struct lldp_agent *agent)
 			}
 		}
 		if (tlv->type == TYPE_4) { /* port description */
-			agent->lldpdu |= RCVD_LLDP_TLV_TYPE4;
-			agent->rx.manifest->portdesc = tlv;
-			tlv_stored = true;
+			if (agent->lldpdu & RCVD_LLDP_TLV_TYPE4) {
+				LLDPAD_INFO("Received multiple port description"
```
orgcandman, line 346: This is a user-visible change in the behavior.  I agree that we should drop the messages, as other implementations do, but there could be deployments with buggy transmitters that will start seeing failures.  We should document this change in the NEWS file.

lldp/rx.c
```diff
@@ -406,21 +406,24 @@ void rxProcessFrame(struct port *port, struct lldp_agent *agent)
...
-
-			if (!err)
-				tlv_stored = true;
-			else if (err == TLV_ERR) {
-				frame_error++;
-				free_unpkd_tlv(tlv);
-				goto out;
-			}
-		}
+                if (!tlv_stored) {
```
orgcandman, line 409: I disagree to this - we should reference count the TLV data, I think.  What if multiple modules reference the same TLV.  For one thing, I'm not confident that there aren't current cases in the codebase where we need this (for example, I don't generally do much testing with DCBX).
Is there a case where we know this currently is happening?  Or is it just a change "to be safe"?

lldp_dcbx.c
```diff
@@ -734,12 +734,16 @@ int dcbx_rchange(struct port *port, struct lldp_agent *agent, struct unpacked_tl
 		if (tlv->info[DCB_OUI_LEN] == DCBX_SUBTYPE2) {
 			if (dcbx->dcbx_st == DCBX_SUBTYPE2)
 				dcbx->manifest->dcbx2 = tlv;
+			else
```
orgcandman, line 737: In this case, perhaps we should introduce a TLV_IGN that we can return?  Then the rxProcessFrame loop will continue processing on the pipeline and if nothing handles it (which we expect in this case) it will free it properly.  WDYT?

lldp/tx.c
rosslagerwall: Yes, I'll fix that too.

lldp/rx.c
rosslagerwall: Sure. I will leave that to you since it looks like `ChangeLog` is normally only updated at release time.

lldp/rx.c
rosslagerwall: If there is a situation currently where multiple modules reference the same TLV then I think it would crash because there is no reference counting so they would both try to free the TLV. Therefore I was updating the code to simply avoid that ever being a possibility.

It's a "to be safe" change that I saw when trying to debug the other memory issues. However, as it is more controversial and I don't have much time or the various setups to investigate it further, I will drop it for now.

lldp_dcbx.c
rosslagerwall: It looks like elsewhere in this function and other parts of the codebase, they use `SUBTYPE_INVALID` as an equivalent of TLV_IGN.

I could use `SUBTYPE_INVALID` here too, but in this case it is not completely ignored - it sets some flags - and the comment above seems to indicate that it intentional to capture them. I don't know why though?

```
		/* Only store a legacy (CIN or CEE) DCBX TLV which matches
		 * the currently configured legacy dcbx mode.
		 * However, capture if any legacy DCBX TLVs are recieved.
		*/
```

rosslagerwall:
Updated as per the comments.

orgcandman:
Thanks!  Merged

PR #91, opened 2023-01-25
Add support for Additional Ethernet Capabilities TLV (Frame Preemption)

vladimiroltean:
The Linux kernel gained support for configuring the IEEE 802.3 MAC Merge layer via ethtool netlink:
https://git.kernel.org/pub/scm/linux/kernel/git/netdev/net-next.git/commit/?id=f3c6e128936e11e62d0af3c52f756194d79cf2e2
The intended usage for this feature is to advertise and negotiate the Frame Preemption feature with the link partner via the Additional Ethernet Capabilities (IEEE 802.3-2018 clause 79.3.7) TLV.
This pull request adds support to lldpad and lldptool for this new TLV, and for talking to the kernel via the ethtool genetlink family. To that end, the lldpad program now links with libnl-genl-3.0 in addition to the libnl-3.0 it used so far.


lldp_8023.c
```diff
@@ -463,6 +464,47 @@ void ieee8023_ifup(char *ifname, struct lldp_agent *agent)
...
+	if (ntoh24(oui) != OUI_IEEE_8023)
+		return SUBTYPE_INVALID;
+
+	subtype = *(tlv->info + OUI_SIZE);
+	switch (subtype) {
+	case LLDP_8023_MACPHY_CONFIG_STATUS:
+	case LLDP_8023_POWER_VIA_MDI:
+	case LLDP_8023_LINK_AGGREGATION:
+	case LLDP_8023_MAXIMUM_FRAME_SIZE:
+		return TLV_OK;
```
orgcandman, line 728: If I understand correctly, we need to keep a reference to the TLV here and release it properly, *or* we need to introduce a new return code (something like TLV_OK_RELEASE) that will tell the rxProcessFrame to delete the TLV object.  Otherwise, I think there could be a memory leak.  Can you confirm?

lldp_ethtool.c
```diff
@@ -0,0 +1,296 @@
...
+	nl_socket_free(sk);
+}
+
+int ethtool_mm_get_state(const char *ifname, struct ethtool_mm *mm)
+{
+	struct nl_sock *sk;
+	struct nl_msg *nlm;
+	int err;
+
+	err = ethtool_genl_start(ETHTOOL_MSG_MM_GET, ETHTOOL_A_MM_HEADER,
```
orgcandman, line 157: Is there a reason to keep creating / deleting the netlink socket?

lldp_8023.c
```diff
@@ -500,6 +699,9 @@ int ieee8023_rchange(struct port *port, struct lldp_agent *agent,
 	case LLDP_8023_LINK_AGGREGATION:
 	case LLDP_8023_MAXIMUM_FRAME_SIZE:
 		return TLV_OK;
+	case LLDP_8023_ADD_ETH_CAPS:
+		return ieee8023_rchange_add_eth_caps(bd, tlv->info + OUI_SUB_SIZE,
```
orgcandman, line 730: Same as before (TLV_OK leak issue).

orgcandman:
Thanks for this - I didn't see a changelog entry for it.  This is a user visible feature, please include a changelog update.

I didn't get a chance to test this code - on kernels which don't support the new API interface, I think it should work okay.  I saw your upstream kernel patch series, and I hope to get some time this week to test.

lldp_ethtool.c
vladimiroltean: I guess I didn't want an API that exposes a stateful struct ethtool_sock *, because that would be required with a reusable socket. I'll try and add it and see how it goes.

vladimiroltean:
Closing this PR and opening another one with the requested changes. I'd like to continue having the code in this form on an immutable branch.

PR #94, opened 2023-03-09
Fixes after #90 and a compile fix

rosslagerwall:


orgcandman:
Thanks for following up - I only did a spot check, and merged.

PR #97, opened 2023-08-02
lldpad: dcbx: prevent null dereference in dcbx_free_data

### LINKED ISSUE #96: lldpad generated core dump with SIGSEGV in dcbx_unregister
Hi,
The lldp on my alma crashed with a core dump.
The analysis gives the following trace
"
Core was generated by `/usr/sbin/lldpad -t'.
Program terminated with signal SIGSEGV, Segmentation fault.
#0 0x000056219c1628ce in dcbx_unregister ()".

we are using :
lldpad-1.0.1-16.git036e314.el8.x86_64
AlmaLinux release 8.6 (Sky Tiger)

Is there a known bug and a fix ?
Thanks,
Srijit

tabraham:
Clearing the TLV with memset clears the list data corrupting the list resulting in a null dereference in dcbx_free_data when encountering such an object.

Fixes #96

### DIFF
--- a/lldp_dcbx.c
+++ b/lldp_dcbx.c
@@ -306,6 +306,10 @@ void dcbx_free_tlv(struct dcbx_tlvs *tlvs)
 	if (tlvs->dcbx2)
 		free_unpkd_tlv(tlvs->dcbx2);
 
+	tlvs->control = tlvs->pg1 = tlvs->pg2 = tlvs->pfc1 =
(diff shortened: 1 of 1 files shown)


penguin359 (approved):
This looks like a valid fix for the crash seen in issue #96. This should prevent the linked-list corruption seen by the user. I have not been able to reproduce that crash myself on any system, but nothing in this patch loose questionable. It should be OK for merge.

orgcandman:
Thanks Thomas, merged.

PR #102, opened 2024-07-03
lldpad: fix using uninitialized value

liuhangbin:
When the length is less than sizeof(cmd.ifname), the cmd.ifname will be printed uninitialized, which could cause information leakage. Fix this by copying the maximum size of ifname before printing it out.

This issue was found through static code analysis.

Fixes: a37b7e0f3b66 ("lldpad: initial git commit")
Fixes: 3b559d8d0b52 ("VDP: vdptool first version")

### DIFF
--- a/lldptool_cmds.c
+++ b/lldptool_cmds.c
@@ -540,7 +540,7 @@ void print_cmd_response(char *ibuf, int status)
 	if (len < sizeof(cmd.ifname)) {
 		memcpy(cmd.ifname, ibuf+CMD_IF, len);
 	} else {
-		printf("Response ifname too long: %*s\n", (int)len, cmd.ifname);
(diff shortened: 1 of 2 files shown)


penguin359 (changes requested):
I don't think the memcpy is the correct fix here as it's shorter than what's in len. Instead, the printf() should probably be changed to print the source value from the memcpy for debug purposes. If the destination is printed, like now, the length parameter should be changed to match sizeof()-1.

liuhangbin:
> Instead, the printf() should probably be changed to print the source value from the memcpy for debug purposes

Thanks, the patch updated.

penguin359 (approved):
This looks good to me and fixes a read from an uninitialized value.

apconole:
Merged - thanks!

PR #103, opened 2024-07-03
lldpad: free the memory allocted by strdup

liuhangbin:
There are some strdup calls that are not freed after use. Let's free them to avoid resource leaks.

This issue was found through static code analysis.

Fixes: b4f9ebc96e0c ("lldpad: implement APP TLV set/get commands")
Fixes: 0313734d05cc ("lldpad: correct IEEE DCBX capabilities check")

### DIFF
--- a/lldp_8021qaz.c
+++ b/lldp_8021qaz.c
@@ -210,6 +210,7 @@ static int read_cfg_file(char *ifname, struct lldp_agent *agent,
 	char arg_path[256];
 	int res = 0, i;
 	int willing, pfc_mask, delay;
+	char *parse = NULL;
 
 	if (agent->type != NEAREST_BRIDGE)
 		return 0;
@@ -354,7 +355,6 @@ static int read_cfg_file(char *ifname, struct lldp_agent *agent,
 
 	/* Read and add APP data to internal lldpad APP ring */
 	for (i = 0; i < MAX_APP_ENTRIES; i++) {
-		char *parse;
 		char *app_tuple;
 		u8 prio, sel;
 		long pid;
@@ -391,8 +391,10 @@ static int read_cfg_file(char *ifname, struct lldp_agent *agent,
 			ieee8021qaz_mod_app(&tlvs->app_head, 0,
 					    prio, sel, (u16) pid, 0);
 		free(parse);
+		parse = NULL;
 	}
 
+	free(parse);
 	return 0;
 }
 
--- a/lldp_8021qaz_cmds.c
+++ b/lldp_8021qaz_cmds.c
@@ -538,9 +538,9 @@ _set_arg_up2tc(struct cmd *cmd, char *args, const char *arg_value,
 	}
 
 	if (test) {
+		free(parse);
 		if (!is_dcbx_hw(cmd->ifname))
 			return cmd_not_capable;
-		free(parse);
 		return cmd_success;
 	}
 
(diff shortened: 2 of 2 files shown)


lldp_8021qaz.c
```diff
@@ -393,6 +393,9 @@ static int read_cfg_file(char *ifname, struct lldp_agent *agent,
 		free(parse);
```
apconole, line 393: You'll need to set parse to NULL here

lldp_8021qaz_cmds.c
```diff
@@ -1191,8 +1192,13 @@ static int get_arg_app(struct cmd *cmd, char *args, UNUSED char *arg_value,
 
 		arg_app_strncat_hw(new_app, hw);
 		strncat(app_buf, new_app, sizeof(app_buf) - strlen(app_buf) - 2);
+
+		free(parse);
```
apconole, line 1196: Same comment - set parse to NULL here.

lldp_8021qaz_cmds.c
```diff
@@ -1191,8 +1192,13 @@ static int get_arg_app(struct cmd *cmd, char *args, UNUSED char *arg_value,
...
 	}
 
+	if (parse)
```
apconole, line 1199: Same comment - the if is not needed.

lldp_8021qaz.c
liuhangbin: Why should we set parse to NULL instead of free it? `strtok` only modified the data.

apconole: I mean, after the free parse will still have a non-zero value.

so then we will double free.  The correct code should look like:

  free(parse);
  parse = NULL;

Does it make sense?

lldp_8021qaz.c
penguin359, line 1: I'm not a big fan of `strtok()` as it is not reentrant or thread-safe and makes the required lifetime of `parse` here unclear. I'd tend to change this to `strtok_r()` instead.

However, that should not be a critical issue for now.

lldp_8021qaz.c
```diff
@@ -210,6 +210,7 @@ static int read_cfg_file(char *ifname, struct lldp_agent *agent,
 	char arg_path[256];
 	int res = 0, i;
 	int willing, pfc_mask, delay;
+	char *parse;
```
penguin359, line 213: It might be good to initialize this to `NULL` in the unlikely case that the loop never reaches `strdup()` during any iterations.

penguin359 (commented):
Once all comments in this PR have been addressed, this looks like a good PR to merge and fix a memory leak. It looks like they would only happen in a case of an incomplete config file where some fields are missing and I wonder if there should be a warning logged, but that is a different issue.

lldp_8021qaz_cmds.c
```diff
@@ -655,9 +655,9 @@ _set_arg_tcbw(struct cmd *cmd, char *args, const char *arg_value,
 		err = cmd_invalid;
 		goto invalid;
 	} else if (test) {
+		free(parse);
```
penguin359, line 658: The last use of `parse` or `strtok()` in this function seems to be the `strtok()` on line 649 inside a `for` loop. This `free()` could be moved to just after the loop removing the need to free it later here and at the end of this function.
```suggestion
    for (i = 0; i < 8 && toked_bw; i++) {
		percent[i] = atoi(toked_bw);
		toked_bw = strtok(NULL, ",");
	}
	free(parse);
```

lldp_8021qaz_cmds.c
penguin359, line 1: I wish GitHub would allow comments on lines outside the PR changes regions.

Returns near lines 939 and 942 also look like early returns that fail to free `parse` before returning. Might be good to add them there or switch to the `reval = ...; goto invalid;` pattern used elsewhere for handling those cases.

lldp_8021qaz_cmds.c
```diff
@@ -1121,6 +1121,7 @@ static int get_arg_app(struct cmd *cmd, char *args, UNUSED char *arg_value,
 	const char *app;
 	u8 prio, sel;
 	int proto, hw = -1, i;
+	char *parse;
```
penguin359, line 1124: Same comment here - initialize parse to `NULL` in case `for` loop never reached `strdup()`.

apconole:
Merged - thanks!

PR #104, opened 2024-07-05
Fix some resource leak issues

liuhangbin:
Fix issues where some resources are allocated but not freed, preventing potential resource leaks.

### DIFF
--- a/lldp_dcbx.c
+++ b/lldp_dcbx.c
@@ -360,8 +360,8 @@ static void dcbx_free_data(struct dcbd_user_data *dud)
 
 struct lldp_module * dcbx_register(void)
 {
-	struct lldp_module *mod;
-	struct dcbd_user_data *dud;
+	struct lldp_module *mod = NULL;
+	struct dcbd_user_data *dud = NULL;
 	int dcbx_version;
 	int i;
 
@@ -385,7 +385,6 @@ struct lldp_module * dcbx_register(void)
 	}
 	dud = malloc(sizeof(*dud));
 	if (!dud) {
-		free(mod);
 		LLDPAD_ERR("failed to malloc LLDP DCBX module user data\n");
 		goto out_err;
 	}
@@ -431,6 +430,8 @@ struct lldp_module * dcbx_register(void)
 	LLDPAD_DBG("%s: dcbx register done\n", __func__);
 	return mod;
 out_err:
+	free(mod);
+	free(dud);
 	LLDPAD_DBG("%s: dcbx register failed\n", __func__);
 	return NULL;
 }
--- a/lldp_dcbx_nl.c
+++ b/lldp_dcbx_nl.c
@@ -279,7 +279,8 @@ static int get_state(char *ifname, __u8 *state)
 		NLMSG_ALIGN(sizeof(struct dcbmsg)));
 
 	if (d->cmd != DCB_CMD_GSTATE) {
-		return -EIO;
+		rval = -EIO;
+		goto out;
 	}
 	if (rta->rta_type != DCB_ATTR_STATE) {
 		rval = -EIO;
@@ -287,8 +288,8 @@ static int get_state(char *ifname, __u8 *state)
 		*state = *(__u8 *)NLA_DATA(rta);
 	}
 
+out:
 	free(nlh);
-
 	return rval;
 }
 
@@ -493,11 +494,15 @@ int get_dcb_capabilities(char *ifname,
 	rta_parent = (struct rtattr *)(((char *)d) +
 		NLMSG_ALIGN(sizeof(struct dcbmsg)));
 
-	if (d->cmd != DCB_CMD_GCAP)
-		return -EIO;
+	if (d->cmd != DCB_CMD_GCAP) {
+		rval = -EIO;
+		goto out;
+	}
 
-	if (rta_parent->rta_type != DCB_ATTR_CAP)
-		return -EIO;
+	if (rta_parent->rta_type != DCB_ATTR_CAP) {
+		rval = -EIO;
+		goto out;
+	}
 
 	rta_child = NLA_DATA(rta_parent);
 	rta_parent = (struct rtattr *)((char *)rta_parent +
@@ -540,6 +545,7 @@ int get_dcb_capabilities(char *ifname,
 	if (rta_parent != rta_child)
 		LLDPAD_DBG("rta pointers are off\n");
 
+out:
 	free(nlh);
 	return rval;
 }
@@ -580,11 +586,15 @@ int get_dcb_numtcs(const char *ifname, u8 *pgtcs, u8 *pfctcs)
 	rta_parent = (struct rtattr *)(((char *)d) +
 		NLMSG_ALIGN(sizeof(struct dcbmsg)));
 
-	if (d->cmd != DCB_CMD_GNUMTCS)
-		return -EIO;
+	if (d->cmd != DCB_CMD_GNUMTCS) {
+		rval = -EIO;
+		goto out;
+	}
 
-	if (rta_parent->rta_type != DCB_ATTR_NUMTCS)
-		return -EIO;
+	if (rta_parent->rta_type != DCB_ATTR_NUMTCS) {
+		rval = -EIO;
+		goto out;
+	}
 
 	rta_child = NLA_DATA(rta_parent);
 	rta_parent = (struct rtattr *)((char *)rta_parent +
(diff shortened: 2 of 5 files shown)


lldp_dcbx_nl.c
```diff
@@ -279,6 +279,7 @@ static int get_state(char *ifname, __u8 *state)
 		NLMSG_ALIGN(sizeof(struct dcbmsg)));
 
 	if (d->cmd != DCB_CMD_GSTATE) {
+		free(nlh);
 		return -EIO;
 	}
```
penguin359, line 284: This is good, but it might be nicer, for code consistency, to make it follow the pattern used in other functions in this file:
```suggestion
	if (d->cmd != DCB_CMD_GSTATE) {
		rval = -EIO;
        goto err_out;
	}
```
And then add the goto label just before the `free(nlh);`

lldp_dcbx.c
```diff
@@ -431,6 +430,8 @@ struct lldp_module * dcbx_register(void)
 	LLDPAD_DBG("%s: dcbx register done\n", __func__);
 	return mod;
 out_err:
+	free(mod);
+	free(dud);
```
penguin359, line 434: You need to initialize `mod` and `dud` to `NULL` in the function prologue as they may be uninitialized under some error code paths that reach here leading to a crash instead of a leak.

qbg/ecp22.c
```diff
@@ -943,6 +943,7 @@ static int ecp22_req2send(char *ifname, unsigned short subtype,
 		goto out;
 	}
 	if (ptlv->size >= ECP22_MAXPAYLOAD_LEN) {
+		free_pkd_tlv(ptlv);
```
penguin359, line 946: This check and the `!ptlv` guard just above this should probably be moved to the top of the file as they are effectively checking the return value of `copy_ptlv(du)` which is the first code actually called in this function. This is also needed to fix the same memory leak in the `!ecp` guard that occurs above which should also call `free_pkd_tlv(ptlv)` on error, but only if `ptlv != NULL` which is only checked for afterwards in the current logic.
```suggestion
	if (!ptlv) {
		rc = -ENOMEM;
		goto out;
	}
	if (ptlv->size >= ECP22_MAXPAYLOAD_LEN) {
		free_pkd_tlv(ptlv);
		rc = -E2BIG;
		goto out;
	}

	eud = find_module_user_data_by_id(&lldp_mod_head, LLDP_MOD_ECP22);
	ecp = find_ecpdata(ifname, eud);
	if (!ecp) {
		free_pkd_tlv(ptlv);
		rc = -ENODEV;
		goto out;
	}
```

lldp_util.c
```diff
@@ -1287,6 +1287,7 @@ int get_arg_val_list(char *ibuf, int ilen, int *ioff,
 			p = (int *) realloc(argvallens,
 				(i/NUM_ARGS + 1) * NUM_ARGS * sizeof(int));
 			if (!p) {
+				free(arglens);
```
penguin359, line 1290: This is good, but it also highlights another possible memory leak. I suspect the `return` just above on line 1282 also needs a call to `free(argvallens)` due to how this loop operates repeatedly calling `realloc()` as needed to ensure there's enough space. It is probably better to instead have the two `free()` calls in the function epilogue and just `numargs = 0; goto out;`. Both `arglens` and `argvallens` will be either `NULL` or the last successfully allocated memory block so it's probably cleanest to clean them both up at the end.

penguin359 (changes requested):
Overall, it looks like this catches several possible memory leaks in the error paths, but it highlights a couple other places that similar changes are needed. Also, care needs to be exercised that this doesn't introduce any worse behaviors like a double-free that can trigger a crash instead.

qbg/ecp.c
```diff
@@ -972,7 +972,7 @@ static void ecp_rx_ProcessFrame(struct vdp_data *vd)
 			free_unpkd_tlv(tlv);
 			vd->ecp.stats.statsTLVsUnrecognizedTotal++;
 		}
-		tlv = NULL;
+		free_unpkd_tlv(tlv);
```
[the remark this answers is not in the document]
liuhangbin: Should I add the `tlv = NULL` in `free_unpkd_tlv`?

lldp_dcbx_nl.c
```diff
@@ -279,16 +279,17 @@ static int get_state(char *ifname, __u8 *state)
...
+		rval = -EIO;
+		goto err_out;
 	}
 	if (rta->rta_type != DCB_ATTR_STATE) {
 		rval = -EIO;
 	} else {
 		*state = *(__u8 *)NLA_DATA(rta);
 	}
 
+err_out:
```
penguin359, line 291: Minor nitpick, as this is both the successful return path as well as the error return path, the label `err_out` is a little misleading. I would just call it `out`, but otherwise, it looks functionally correct.

This convention is already used elsewhere in the code.

lldp_dcbx_nl.c
```diff
@@ -540,6 +545,7 @@ int get_dcb_capabilities(char *ifname,
 	if (rta_parent != rta_child)
 		LLDPAD_DBG("rta pointers are off\n");
 
+err_out:
```
penguin359, line 548: Same comment here.

lldp_dcbx_nl.c
```diff
@@ -618,9 +628,10 @@ int get_dcb_numtcs(const char *ifname, u8 *pgtcs, u8 *pfctcs)
 	if (rta_parent != rta_child)
 		LLDPAD_DBG("rta pointers are off\n");
 
-	free(nlh);
 	if (found != 3)
 		rval = -EIO;
+err_out:
```
penguin359, line 633: And here as well.

lldp_util.c
```diff
@@ -1311,21 +1311,21 @@ int get_arg_val_list(char *ibuf, int ilen, int *ioff,
...
+			goto err_out;
 		}
 	}
 	numargs = i;
 	for (i = 0; i < numargs; i++) {
 		args[i][*(arglens+i)] = '\0';
 		argvals[i][*(argvallens+i)] = '\0';
 	}
+
+err_out:
```
penguin359, line 1328: And here.

qbg/ecp.c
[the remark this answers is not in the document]
penguin359: tlv can't be modified inside of `free_unpkd_tlv` as it's passed by value, but it is a good idea to set it to `NULL` after calling that function which is already how the code is written now so no changes needed here.

qbg/ecp22.c
```diff
@@ -958,6 +958,7 @@ static int ecp22_req2send(char *ifname, unsigned short subtype,
 	ecp22_add_payload(ecp, payda);
 out:
 	LLDPAD_DBG("%s:%s rc:%d\n", __func__, ifname, rc);
+	free_pkd_tlv(ptlv);
```
penguin359, line 961: I don't believe this is the correct place to free this data. The ownership of this structure is passed to `payda` which is added to a list in the ecp data and remains in existence after this function. The TLV should only be freed in error return paths as it was originally. It was missing from some error pathways, but the current code is likely to lead to a crash now.

An alternative to removing this call to `free_pkg_tlv()`, you can set `ptlv` to `NULL` right after it's assigned to `payda->ptlv` on line 955 like this:
```
payda->ptlv = ptlv;
ptlv = NULL;
```
As I am not completely familiar with this area of the code, please report back if you believe this is in error and what you believe the actual lifetime of `ptlv` is.

penguin359 (changes requested):
This PR appears to fix several memory leaks that occur in the error return paths of multiple functions and should be merged once all issues are resolved, but there does appear to be one critical blocker that must be resolved first. There is one memory free that is on the successful code path for a data structure that appears to have a lifetime beyond this function call. Please review the comment and apply an appropriate fix as needed.

qbg/ecp22.c
liuhangbin: Yes, you are right, I should not call free_pkd_tlv() in the end if `rc == 0`. I think we can return before `out` if rc is 0.

lldp_util.c
penguin359: Small nitpick, but this one is still labeled as err_out on the common return path.

penguin359 (approved):

penguin359 (approved):
This PR fixes several memory leaks and I don't see anymore blocking issues that prevent this from being merged at this time. It is also passing the full CI build suite with no failures:

https://github.com/penguin359/openlldp/actions/runs/10368328451

lldp_util.c
liuhangbin: Ah, I didn't notice this is for lldp_utils.c. Sorry. Fixed now.

penguin359 (approved):
All concerns seem to have been addressed and I see no further issues remaining. This PR should fix a number of potential memory leaks that exist in OpenLLDP.

apconole:
Applied, thanks.  I will dedicate some time next week to backport these to 1.1 branch.

PR #106, opened 2024-08-02
lldpad: using inet_ntop

liuhangbin:
Function inet_ntoa is deprecated. Use inet_ntop instead. Also convert inet_aton to inet_pton.

### DIFF
--- a/qbg/vdp22cisco_oui.c
+++ b/qbg/vdp22cisco_oui.c
@@ -130,13 +130,13 @@ bool cisco_str2vdpnl_hndlr(struct vdpnl_oui_data_s *vdp_oui_p, char *token)
 				return false;
 			}
 			strncpy(v4_addr_str, token, data_len);
-			ret = inet_aton(v4_addr_str,
+			ret = inet_pton(AF_INET, v4_addr_str,
 					&vdp_cisco_oui_p->l3_addr.ipv4_address);
 			LLDPAD_DBG("V4adr %s 0x%lx\n", v4_addr_str,
 				   (unsigned long)
 				  vdp_cisco_oui_p->l3_addr.ipv4_address.s_addr);
 			free(v4_addr_str);
-			if (!ret) {
+			if (ret <= 0) {
 				LLDPAD_ERR("%s: Incorrect addr\n", __func__);
 				return false;
 			}
--- a/configure.ac
+++ b/configure.ac
@@ -78,7 +78,7 @@ AC_CHECK_LIB(nl3, rtnl_link_get_by_name)
 
 AC_CHECK_FUNCS([alarm])
 AC_CHECK_FUNCS([gettimeofday])
-AC_CHECK_FUNCS([inet_ntoa])
+AC_CHECK_FUNCS([inet_ntop])
 AC_CHECK_FUNCS([localtime_r])
 AC_CHECK_FUNCS([memmove])
 AC_CHECK_FUNCS([memset])
--- a/lldp_util.c
+++ b/lldp_util.c
@@ -1092,7 +1092,8 @@ int get_ipaddrstr(const char *ifname, char *ipaddr, size_t size)
 	rc = get_saddr(ifname, &sa);
 	if (rc == 0) {
 		memset(ipaddr, 0, size);
-		strncpy(ipaddr, inet_ntoa(sa.sin_addr), size);
+		if (inet_ntop(AF_INET, &sa.sin_addr, ipaddr, size) == NULL)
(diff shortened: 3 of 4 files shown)


apconole:
LGTM - I'll merge on Tuesday (unless someone flags an issue).

qbg/vdp22cisco_oui.c
```diff
@@ -130,7 +130,7 @@ bool cisco_str2vdpnl_hndlr(struct vdpnl_oui_data_s *vdp_oui_p, char *token)
 				return false;
 			}
 			strncpy(v4_addr_str, token, data_len);
-			ret = inet_aton(v4_addr_str,
+			ret = inet_pton(AF_INET, v4_addr_str,
```
penguin359, line 133: `inet_pton()` does exhibit different behaviors than `inet_aton()` which should be examined. For one, where `inet_aton()` will return 1 for success and 0 for malformed address, `inet_pton()` has a third, possible return value of -1 on other error. The only documented error is if an invalid address family is specified and I don't see any other possible cases, but it still might be better to change the check from `if(!ret)` to something more like `if(ret <= 0)` just to future-proof the code.

The second behavior is that `inet_aton()` supports older address formats of A, A.B, and A.B.C, the first one being the case where the IPv4 address is provided as a single 32-bit number instead of the standard dotted-octet format. `inet_pton()` will return a zero for malformed address on these other cases. I don't think this should be an issue, but it should be considered in case there are other side-effects. `inet_aton()` is also not marked as deprecated like `inet_ntoa()` is, but it is nice to be consistent.

penguin359 (changes requested):
There is one required code change that will trigger a build error under some C compilers and should be fixed before merge. A few other notes are included, but nothing else I see that would block this PR.

qbg/vdp22cisco_oui.c
liuhangbin: > The second behavior is that `inet_aton()` supports older address formats of A, A.B, and A.B.C, the first one being the case where the IPv4 address is provided as a single 32-bit number instead of the standard dotted-octet format. `inet_pton()` will return a zero for malformed address on these other cases. I don't think this should be an issue, but it should be considered in case there are other side-effects. `inet_aton()` is also not marked as deprecated like `inet_ntoa()` is, but it is nice to be consistent.

Hmm, is the `token` an ipv4 dot string or a number? Because I saw `oui_vdp_hexstr2bin` just make the hex to bin and I can't find where the `.` added.

penguin359: I haven't taken the time yet to look at the code in more detail to see if it ever would come in as a number, but most likely it will always be in dotted-decimal form and there's no issue.

My overall concern with this PR is just making sure we don't break any existing behavior with changes that offer little to no gain. While `inet_ntoa()` is marked as deprecated, it's more to discourage future developers using it and making migrations to IPv6 more difficult in the future. However, it's not going to go away or break it's behavior anytime in the near or mid future. The function `inet_aton()` isn't even marked as deprecated, possibly because it has behaviors not covered in the newer functions. The use of `inet_ntop()` and `inet_pton()` over their predecessors only offer a real benefit when they are not hard-coded to `AF_INET`. Otherwise, they behave essentially the same (or close to it). I just want to make sure we aren't accepting in patches that don't ultimately improve things and actually end up breaking something in some less-often tested corner case.

liuhangbin: Fair enough. I'm OK to drop this PR. Thanks.

penguin359 (approved):
Thank you for applying that change. There were no build errors from the latest CI run as seen here:

https://github.com/penguin359/openlldp/actions/runs/10280156633

The compiler regression is a good example of why I am scrutinizing the PR so heavily. Small changes can have unexpected side-effects if not reviewed fully and loosing support for older compilers just to move to a newer version of a function that does the same thing is not a good trade-off, but that issue is now resolved and no longer blocking this PR.

Another note on the `inet_pton()` change. That is in a file supporting functionality contributed by Cisco which was added nearly 10 years ago and untouched until now. It's for supporting a feature called VDP which I am unfamiliar with and is kind of in the "don't fix what ain't broke category" and as far as I know, still works as well as it did 10 years ago. In any case, I see the usage is documented in the Git commit message and it's expected to take dotted-decimal notation as seen here:

-c oui=cisco,ipv4_addr=a.b.c.d

I don't see any regressions in the current code and it should be fine to merge. It's now up to @apconole to decide what to do with it.

apconole:
Merged.

PR #107, opened 2024-08-11
Enable testing on newer GCC releases and clean up warnings on them

penguin359:
In my earlier PR for enabling build testing automation, I stopped at GCC 9.x because newer releases were triggering warnings and would therefore fail the build since `-Werror` was in effect. This PR adds in those versions and fixes all warnings up through GCC 14.x.

The main issue is the call to `strncpy()` in the `STRNCPY_TERMINATED` macro triggers GCC's concerns over a possible non-terminated string even though a quick analysis will show that it is indeed terminated. This warning is present up through at least GCC 14.x. I initially tried to replace the macro with a call to `strlcpy()` which is a safe version of `strncpy()`, but that is too new for many systems so I have the warning silenced inside the macro on newer GCC versions. There are no more warnings on any version of GCC tested now.

* Run all matrix combinations even if some fail
* Move to a non-deprecated checkout version
* Added newer GCC versions for test
* Added missing build dependencies
* Use the safe `strlcpy()` when present
* Make macro safer when `strlcpy()` is not available
* Disable invalid warning about use of `strncpy()`


penguin359:
Hmm, I guess since I am a contributor outside the project modifying the GitHub workflow, it won't automatically run the workflow without approval. I suppose that makes sense, but I was expecting it would just run. In any case, is pretty quick to run the checks once approved.

penguin359:
After reviewing the build logs from the `-Wstringop-truncation` warning I tried to silent in my commit ede7ee3258d219cc7460731c4097c509f60615cd, I see that it was actually trying to warn me about the issue with the fixed-sized string string arrays where the source was `IFNAMSIZ+1` and the destination was only `IFNAMSIZ`. That commit should probably be backed out and the warning should go away once issue #108 is fixed. Interestingly, while GCC 10.x seems to detect this with just a `-Wall`, it looks like GCC 9.x can trigger the warning, but only with ASAN enabled and using the fortified `strncpy()`. The build logs showing the warning are here:

https://github.com/penguin359/openlldp/actions/runs/10337712563/job/28614883097#step:8:153

apconole:
Applied, thanks!

PR #110, opened 2024-08-15
lldpad: remove device from config file when ports are removed

liuhangbin:
When ports are removed, there is no need to keep them in the configuration file. Otherwise, the port list could grow too large to manage. For example, if a user adds hundreds of virtual ports for testing and then removes them after the test, the configuration file would retain this information indefinitely. Removing these entries ensures the configuration file remains manageable.


apconole:
This is a great enhancement.  Thanks Hangbin - merged!

PR #112, opened 2024-10-09
lldpad: do not remove config when stopping lldpad

liuhangbin:
When lldpad is terminated by systemd or kill, the remove_all_adapters() function is called. However, this function is intended to release port resources, and the configuration file should not be destroyed. We only need to remove the port when it is explicitly deleted. Otherwise, all configurations will be lost when lldpad is stopped.


apconole (approved):
LGTM - thanks

apconole:
Applied.

PR #116, opened 2024-11-25
treewide: drop all mentions of open-lldp.org

### LINKED ISSUE #115: open-lldp.org redirects to realdecatorce.net
As far as i can tell open-lldp.org is the official project domain. It's mentioned in a few places in the repo. Hoewer, i'm not sure wheter it's still under project control.

When requesting open-lldp.org there are a bunch of redirects and in the end you land on `www.realdecatorce.net`

```
$ curl -s -v --stderr - -L https://www.open-lldp.org/ | grep -i "location:"
< location: https://better-explorer.com/
< location: https://hello88v.net/
< location: https://www.zoomaru5.com/
< location: https://hello88v.org/
< location: https://www.idrawdigital.com/
< location: https://www.ingoodcompanyworkplaces.com/
< location: https://www.realdecatorce.net/
```

And that page doesn't really look like it's related to openlldp:

image

Looking at [archive.org](https://web.archive.org/web/20240000000000*/https://www.open-lldp.org/) this probably happened at some point after 2024-01-20.

Last successful archive.org capture: https://web.archive.org/web/20240120045633/http://open-lldp.org/

If control over the domain is permanently lost, it would probably make sense to remove all mentions of the domain from the repo.

([gist with full curl following all redirects](https://gist.github.com/herbetom/5e43cf28fb69847e7f83fad8f49e3245))

herbetom:
Just an idea to replace it with the GitHub Issue Tracker.

closes #115

### DIFF
--- a/README
+++ b/README
@@ -564,5 +564,5 @@ Support
 =======
 
   Contact Information:
-  LLDP-devel Mailing List <lldp-devel@open-lldp.org>
+  Issue Tracker: https://github.com/intel/openlldp/issues
   Intel Corporation, 5200 N.E. Elam Young Parkway, Hillsboro, OR 97124-6497
--- a/clif.c
+++ b/clif.c
@@ -25,7 +25,7 @@
   the file called "COPYING".
 
   Contact Information:
-  open-lldp Mailing List <lldp-devel@open-lldp.org>
+  Issue Tracker: https://github.com/intel/openlldp/issues
 
 *******************************************************************************/
 
--- a/config.c
+++ b/config.c
@@ -20,7 +20,7 @@
   the file called "COPYING".
 
   Contact Information:
-  open-lldp Mailing List <lldp-devel@open-lldp.org>
+  Issue Tracker: https://github.com/intel/openlldp/issues
 
 *******************************************************************************/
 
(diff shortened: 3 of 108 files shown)


penguin359 (approved):
Everything here looks good and URL works. No issues with merging it as is, however, the long-term home might be moving in the near future. It should not be critical as the current address should be a semi-permanent redirect to the new home, but we'd still want to update it if the move becomes permanent.

penguin359:
Hmm, I am unable to assign or request a review from @apconole on this PR.

apconole:
@penguin359 it seems like I no longer have write access to the repository.

penguin359:
@apconole Well, that explains why I couldn't mention you earlier. I suspect that you were originally added manually with the Web UI and that when they updated the permissions recently, dropping owners who are no longer with Intel, that it reset all the permissions to match what they maintain internally. Unfortunately, I only seem to be a maintainer, not an owner at the moment so I can't add you back in. I'll have to get in touch with one of the remaining owners and should probably get myself included in that list.

An owner should be able to re-add you from the Web UI, but it is at risk of getting dropped again. I'll see if there's a way to add an external user to the internal copy of the permissions, but I'm not sure if there is an official way to do that.

apconole (approved):

PR #117, opened 2024-12-06
correct error version

Shenzhongwei001:


apconole (commented):
I don't think this change is appropriate.  We have the 1.1 branch which carries the version for 1.1.  This branch is the current mainline, and that will become 1.2 whenever enough features have been accumulated.

configure.ac
```diff
@@ -1,4 +1,4 @@
-AC_INIT([lldpad], [1.2.99], [lldp-devel@open-lldp.org])
+AC_INIT([lldpad], [1.1.1], [lldp-devel@open-lldp.org])
```
apconole, line 1: Why do you think this should change?

This value will be a hint on bug reports.  IE: if someone writes a bug and shows 1.2.90, then we know it is 
against latest main branch.  OTOH, 1.1.1 should be found on branch-1.1 and that will be a very specific version.

apconole:
Closing this.  If you're looking for the 1.1 branch, see the branch-1.1 

Thanks.

Issue #118, opened 2024-12-23
coredump in lldpad


huyizhen:
I have an environment with 500+ virtual NICs, and core dumps occur when I run lldpad on it.
When the function eloop_sock_table_dispatch calls FD_ISSET, the value of file descriptor table->table[i].sock exceeds 1024 (the kernel fds structure has only 1024 bits). As a result, the glibc determines that a buffer overflow occurs and aborts the process.
Have you ever had this problem? Other than recompiling the kernel and glibc magnifying the size of fds, is there any other way around this problem?
Change select to poll may solve this problem. Do you have any plans to change it?

penguin359:
This is a bug and it would be best to move on to using `poll()`/`epoll()` in the future. In the meantime, I will patch this to check for potential overflow.

huyizhen:
Thanks for your reply and pull request [penguin359](https://github.com/penguin359)!
I recently tried to change select to poll. While ensuring that the original problem was resolved, I also do some tests, including lldpad -xxx, lldptool set-lldp xxx and memory leak test.
I also plan to test the NICs online/offline and the LLDP protocol interaction.
Welcome to review my PR and provide feedback, i will also consider your pr to solve my problem as a temp plan.
Best regards!

PR #119, opened 2024-12-27
Protect uses of select() from crash on too many file descriptors

penguin359:
The fd_set macros don't check for overflow and will cause an
out-of-bounds access when given a file description >= FD_SETSIZE (1024)

This is to address issue #118 which originally reported this crash. Ultimately, we should switch to using either `poll()` or `epoll()` for handling file descriptor events, but this PR will at least prevent `lldpad` from a crash when used with a large (> ~500) interfaces. Some interfaces will fail to operate correctly, but this at least allows for partial functionality until the full fix can be included.

### DIFF
--- a/eloop.c
+++ b/eloop.c
@@ -180,6 +180,17 @@ static void eloop_sock_table_remove_sock(struct eloop_sock_table *table,
 }
 
 
+static inline void warn_too_many_fds()
+{
+	static bool warned = false;
+
+	if (!warned) {
+		fprintf(stderr, "unable to handle more than %d file descriptors\n", FD_SETSIZE);
+		fprintf(stderr, "too many interfaces being watched\n");
+		warned = true;
+	}
+}
+
 static void eloop_sock_table_set_fds(struct eloop_sock_table *table,
 				     fd_set *fds)
 {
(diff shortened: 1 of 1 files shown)


penguin359 (commented):
Hmm, seems I am still unable to assign this to @apconole for a review.

eloop.c
```diff
@@ -571,6 +591,11 @@ void eloop_wait_for_read_sock(int sock)
 	if (sock < 0)
 		return;
 
+	if (sock >= FD_SETSIZE) {
+		warn_too_many_fds();
```
apconole, line 595: I'm okay with this being added, but a quick grep through the code doesn't seem to find any in-tree callers for eloop_wait_for_read_sock.

Maybe we should delete it instead?

penguin359: Yes, I only see `eloop_register_read_sock()`. and `eloop_unregister_read_sock()` used, but not `eloop_wait_for_read_sock()`. Also, it looks like `eloop_register_sock()`, `eloop_unregister_sock()`, and `os_get_time()` can all be made static.

penguin359: Actually, I found several things to clean-up in eloop.c/eloop.h including function prototypes in the header files that refer to functions that have never existed in the Git history. I'll save that clean-up for another PR, but I think I can drop that function for now.

apconole: If you are okay with it, I'll merge this patch and backport it to the 1.1 branch.

apconole:
Merged and backported.  Thanks!

PR #120, opened 2024-12-27
lldpad eloop select change to poll

huyizhen:
In an environment with 576 virtual NICs, when the function eloop_sock_table_dispatch calls FD_ISSET, the value of file descriptor table->table[i].sock exceeds 1024 (the kernel fds structure has only 1024 bits). As a result, the glibc determines that a buffer overflow occurs and aborts the process.

To solve this problem, we change select to poll because poll has no restriction on the fd size. Poll allows file descriptors to listen to different events, such as POLLIN and POLLOUT. Therefore, no need to create different queues for different events, we delete writers and exceptions and rename readers to sock_table.


penguin359:
@huyizhen Please fix the build failures and push an update.

huyizhen:
> @huyizhen Please fix the build failures and push an update.

Fixed

penguin359 (changes requested):
There are some issues with how the list of `pollfd`s are being constructed and passed to `poll()`. There should be one entry in the array for each requested fd, but no more. The current logic creates one entry for every fd up to max used so it will have entries for stdin, stderr, and other fds that are not desired. I also don't think those slots are being initialized after realloc if they don't map to a specific fd that is in the sock table.

eloop.c
```diff
@@ -468,37 +423,31 @@ int eloop_register_signal_reconfig(eloop_signal_handler handler,
...
 
-		eloop_sock_table_set_fds(&eloop.readers, rfds);
-		eloop_sock_table_set_fds(&eloop.writers, wfds);
-		eloop_sock_table_set_fds(&eloop.exceptions, efds);
-		res = select(eloop.max_sock + 1, rfds, wfds, efds,
-			     eloop.timeout ? &_tv : NULL);
+		fds = (struct pollfd *)
+			realloc(fds, (eloop.max_sockfd + 1) * sizeof(struct eloop_sock));
+		if (fds == NULL) {
+			perror("realloc");
```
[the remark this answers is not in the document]
huyizhen: `eloop.sock_table.count` may be changed, so realloc should be used here.
`perror("realloc");` -> `perror("eloop_run realloc");`

huyizhen:
Great! I have modified the code according to your issues. Please review the latest code again. Thank you.
How to trigger the build? Since I found that after `git push - f`, the build was not automatically triggered.

huyizhen:
Dear penguin359, is there anything else I need to do to merge this PR?@penguin359

penguin359:
This PR touches quite a few lines of code and now it is modifying 4 different files. It will take a little time to review and some maintainers are currently still out for the holidays.

huyizhen:
Thanks for your reply, I will wait patiently for the review, happy holidays!

eloop.c
```diff
@@ -148,24 +146,22 @@ static int eloop_sock_table_add_sock(struct eloop_sock_table *table,
...
-		eloop.max_sock = sock;
 	table->changed = 1;
 
 	return 0;
 }
 
 
 static void eloop_sock_table_remove_sock(struct eloop_sock_table *table,
-                                         int sock)
+                                         int sockfd)
```
apconole, line 156: Was there a reason to change this name?  IIUC, sock is always an fd.

eloop.c
```diff
@@ -181,22 +177,20 @@ static void eloop_sock_table_remove_sock(struct eloop_sock_table *table,
 
 
 static void eloop_sock_table_set_fds(struct eloop_sock_table *table,
-				     fd_set *fds)
+                                     struct pollfd *fds)
```
apconole, line 180: This is mixing tabs and whitespaces.  Please use the tabs for this file (general comment applies everywhere).

eloop.c
```diff
@@ -238,56 +231,29 @@ static void eloop_sock_table_destroy(struct eloop_sock_table *table)
...
-	case EVENT_TYPE_WRITE:
-		return &eloop.writers;
-	case EVENT_TYPE_EXCEPTION:
-		return &eloop.exceptions;
-	}
-
-	return NULL;
+	struct pollfd sock = {0};
+	sock.fd = sockfd;
+	sock.events = POLLIN;
```
apconole, line 239: I don't understand, we previously had checks for ReadyToRead and ReadyToWrite, but your change is only handling ReadyToRead.

eloop.c
```diff
@@ -522,16 +481,11 @@ void eloop_run(void)
...
-		eloop_sock_table_dispatch(&eloop.writers, wfds);
-		eloop_sock_table_dispatch(&eloop.exceptions, efds);
+		eloop_sock_table_dispatch(&eloop.sock_table, fds, POLLIN);
 	}
-
 out:
-	free(rfds);
-	free(wfds);
-	free(efds);
+	if (fds)
```
apconole, line 487: No need for this if() check.  Just free(fds).

include/eloop.h
```diff
@@ -22,31 +22,20 @@
 
 #ifndef ELOOP_H
 #define ELOOP_H
+#include <poll.h>
```
apconole, line 25: This should go in the .c file rather than the this header.

apconole:
I've added some comments.  I think this needs to be broken into a few commits.  There are renames and other types of cleanups interspersed with the actual meat of the change.  Please do your best to organize the changes so that each commit will be a separate logical change.

eloop.c
huyizhen: To distinguish between struct pollfd and int, I named struct pollfd sock and int sockfd. I'll now name both struct pollfd and int fd sock.

eloop.c
huyizhen: The previous code mixed tabs and spaces, and I chose spaces before. I'm going to use tabs all for indentation now.

eloop.c
huyizhen: The previous code implements only eloop_register_read_sock and eloop_unregister_read_sock, but does not provide the write interface. Actually, other codes do not create sock to write. Therefore, only ReadyToRead is implemented.

eloop.c
huyizhen: If the while loop is not entered, fds will be null, and releasing a null pointer will cause a crash, so I think it is necessary to determine whether fds is empty.

eloop.c
```diff
@@ -564,16 +516,16 @@ int eloop_terminated(void)
 }
 
 
-void eloop_wait_for_read_sock(int sock)
+void eloop_wait_for_read_sock(int sockfd)
```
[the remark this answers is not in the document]
huyizhen: Yes, I can submit a separate commit to removes eloop_wait_for_read_sock after the PR is received.

include/eloop.h
huyizhen: If poll.h is placed in a .c file, eloop.h is no longer self-contained. All .c files that contain eloop.h must contain poll.h.
1) Put poll.h into eloop.h so that eloop.h is self-contained.
2) Place poll.h in .c files that contain eloop.h before eloop.c and config.c.
Which do you think is better?

eloop.c
```diff
@@ -218,15 +212,14 @@ static void eloop_sock_table_dispatch(struct eloop_sock_table *table,
 
 static void eloop_sock_table_destroy(struct eloop_sock_table *table)
 {
-	int rc, tc, sock;
-
-	if (table) {
+	int rc, tc, sockfd;
```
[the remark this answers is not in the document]
huyizhen: Fixed

eloop.c
```diff
@@ -67,9 +67,11 @@ int os_get_time(struct os_time *t)
 	} \
 } while (0)
 
+#define os_time_to_ms(a) \
```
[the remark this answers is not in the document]
huyizhen: Fixed

huyizhen:
> I've added some comments. I think this needs to be broken into a few commits. There are renames and other types of cleanups interspersed with the actual meat of the change. Please do your best to organize the changes so that each commit will be a separate logical change.

Thank you for your comments, I've changed the code as much as possible, but there are a few more comments, including free(fds), the location of poll.h, ReadyToWrite, and I'd like to discuss it with you further.
Switching from select to poll does bring a lot of changes, and I've reverted the previous code as much as possible to make the patch easier to read.
I think the logic in the current patch is all necessary. In other words, if the separation continues, it will be difficult for future readers to understand. For example, retaining some FD_XX interfaces does not affect the commit, but people may have doubts when reading this patch, why the author didn't remove all FD_XX interfaces?
Do you have a better way or suggestion for separating commit?
Looking forward to your next review, thank you!

eloop.c
penguin359: Freeing a non-valid pointer leads to a crash, but `NULL` is a supported code path and becomes effectively a no-op. Setting a pointer to `NULL` right after a `free()` is therefore a common approach, but a guard on `free()` shouldn't be needed in this paradigm.

eloop.c
penguin359: Those renames add quite a bit of noise and make the review harder. It looked like most of the header file was just thar rename which doesn't affect the functionality. Keeping code changes smaller helps when working with a first-time contributor.

huyizhen: yeah, I have already revert the rename of sock, the number of modified lines in eloop.h is reduced from 52 to 23.

huyizhen:
Dear apconole and penguin359, could you please continue to review this PR? I have resolved or replied to the current issues, thank you! @apconole @penguin359

penguin359:
One of my concerns with this PR that I saw in my initial assessment was that it collapsed the event handler into being indexed only by file descriptor instead of file descriptor and event type. That means that it could overwrite a different handler waiting on the file descriptor for a different event. The last handler to wait would win. However, after taking a detailed look through the original code, I see that neither `EVENT_TYPE_WRITE` nor `EVENT_TYPE_EXCEPTION` has ever been used by any handlers anywhere in Git history, only `EVENT_TYPE_READ`. That could be considered over-engineering and there's no current need to keep it. I still want to further review it, but that concern has been alleviated.

huyizhen:
> One of my concerns with this PR that I saw in my initial assessment was that it collapsed the event handler into being indexed only by file descriptor instead of file descriptor and event type. That means that it could overwrite a different handler waiting on the file descriptor for a different event. The last handler to wait would win. However, after taking a detailed look through the original code, I see that neither `EVENT_TYPE_WRITE` nor `EVENT_TYPE_EXCEPTION` has ever been used by any handlers anywhere in Git history, only `EVENT_TYPE_READ`. That could be considered over-engineering and there's no current need to keep it. I still want to further review it, but that concern has been alleviated.

Thank you for your comments, looking forward to your review!

eloop.c
```diff
@@ -181,22 +174,19 @@ static void eloop_sock_table_remove_sock(struct eloop_sock_table *table,
 
 
 static void eloop_sock_table_set_fds(struct eloop_sock_table *table,
-				     fd_set *fds)
+				     struct pollfd *fds)
 {
 	int i;
-
-	FD_ZERO(fds);
-
```
penguin359, line 189: There are a number of places where whitespace or blank lines has been stripped out. I would caution you to be careful to try and preserve the original code formatting when that is not the primary purpose for the code change. In particular, I generall prefer a blank line after variable declarations before the first statement of the block.

eloop.c
```diff
@@ -522,16 +481,11 @@ void eloop_run(void)
 
 		if (res <= 0)
 			continue;
-
-		eloop_sock_table_dispatch(&eloop.readers, rfds);
-		eloop_sock_table_dispatch(&eloop.writers, wfds);
-		eloop_sock_table_dispatch(&eloop.exceptions, efds);
+		eloop_sock_table_dispatch(&eloop.sock_table, fds, POLLIN);
```
penguin359, line 484: Hardcoding `POLLIN` here feels funny, but it is faithful to the original code where specific tables were passed in and keeps the same logic flow in place. I don't really have a recommendation so we can keep this as is.

eloop.c
```diff
@@ -241,53 +230,26 @@ static void eloop_sock_table_destroy(struct eloop_sock_table *table)
...
+	struct pollfd pfd = {0};
+	pfd.fd = sock;
+	pfd.events = POLLIN;
+	return eloop_register_sock(pfd, handler, eloop_data, user_data);
 }
 
-
-void eloop_unregister_read_sock(int sock)
+int eloop_register_sock(struct pollfd pfd,
+						eloop_sock_handler handler,
```
penguin359, line 240: The source code is formatted with 8-space tabs so this change in indentation causes these parameters to be halfway across my screen.

eloop.c
```diff
@@ -219,11 +209,10 @@ static void eloop_sock_table_dispatch(struct eloop_sock_table *table,
 static void eloop_sock_table_destroy(struct eloop_sock_table *table)
 {
 	int rc, tc, sock;
-
-	if (table) {
+	if (table->table) {
```
penguin359, line 212: This is definitely a better guard than `table` as only `table->table` is dynamically allocated. 👍

eloop.c
```diff
@@ -181,22 +174,19 @@ static void eloop_sock_table_remove_sock(struct eloop_sock_table *table,
 
 
 static void eloop_sock_table_set_fds(struct eloop_sock_table *table,
-				     fd_set *fds)
+				     struct pollfd *fds)
```
penguin359, line 177: While I appreciate sticking to the original structure of the code and trying to keep changes to a minimal, this function is probably one that should be refactored as it's just awkward here with how it's used. It was difficult to understand at first glance with the new code. AS it's written now, it is correct, but once this PR is merged in, we might want to look to clean up several functions as the whole notion of having to recreate the file descriptor list passed in to `poll()` is just a relic of `select()` that would destroy the list on each call. No action is needed from you on this right now.

penguin359 (changes requested):
Thank you for taking the time to work on this.

After a thorough review, the code looks functionally correct. I do have a few small changes I've recommended in code layout, but no changes needed to the logic.

eloop.c
huyizhen: Haha, I'm using a tab length of 4, so a lot of function arguments and function names are not aligned, I'm puzzled, and now I finally know why. After changing the tab length, the code now aligns very well.

eloop.c
huyizhen: Yes, but note that eloop_sock_table_dispatch may modify eloop.sock_table. Perhaps we can make fds a member of eloop.sock_table and modify it dynamically in eloop_register_sockhe and eloop_unregister_sock.

eloop.c
huyizhen: I've always thought that free null pointers may cause undefined behavior such as crash, but I just checked the C standard and experimented to confirm that free null pointers is no-op, thank you for noting me that, I've removed the if statement.

penguin359:
I'll say it's looking pretty good right now and seems to be easy to follow the changes. I'll try testing this out and add a formal approval if it all is working.

apconole:
@penguin359 were you planning on acking this?

penguin359:
Yes, let me take one last look at the lastest version.

huyizhen:
> Yes, let me take one last look at the lastest version.

I will resolve the conflict today.

huyizhen:
@penguin359 
Hello. I looked at the conflicting content, and if I want this PR be merged, I would roll back most code of https://github.com/intel/openlldp/pull/119. 
What's your idea? If you think the PR-119 solution is sufficient to solve this problem, then the current PR can be closed.

huyizhen:
@penguin359 I found that you added commit 734561a, and then this PR no longer reported conflicts. But I don't see the file changes of 734561a, and I guess 734561a is a reverted version of RP-119, right?
So, is there anything else I need to do to merge this PR?

penguin359:
Yes, the commit effectively reverts #119 for the overlapping code to resolve the conflict and brings this PR in line with the latest `master` branch. I will begin reviewing the full change shortly. Nothing needed from you at this time.

apconole:
can you add your Signed-off-by: line?  @huyizhen - I haven't done the review, but if you post a signed-off-by line here, I can add it on commit, assuming @penguin359 doesn't have any further comments.

huyizhen:
> can you add your Signed-off-by: line? @huyizhen - I haven't done the review, but if you post a signed-off-by line here, I can add it on commit, assuming @penguin359 doesn't have any further comments.

Thanks, I have added Signed-off-by.

eloop.c
```diff
@@ -484,43 +449,39 @@ int eloop_register_signal_reconfig(eloop_signal_handler handler,
...
 				os_time_sub(&eloop.timeout->time, &now, &tv);
 			else
 				tv.sec = tv.usec = 0;
-			_tv.tv_sec = tv.sec;
-			_tv.tv_usec = tv.usec;
+			timeout = os_time_to_ms(&tv);
+		}
+
+		fds = (struct pollfd *)
+			realloc(fds, eloop.sock_table.count * sizeof(struct eloop_sock));
```
penguin359, line 475: I believe we are slightly over-allocating memory here. We only need `count * sizeof(struct pollfd)`, not the full `struct eloop_sock` for the `realloc()`.

penguin359 (changes requested):
Only saw one small issue with the latest update. Not a complete show-stopper, but I believe we are allocating a little more memory in the event loop than is required. Otherwise, it looks pretty good.

eloop.c
apconole: We can adjust this on the merge if @huyizhen agrees.

apconole:
I think I've found some issues in here while doing a final check, but I will push a version tomorrow.

apconole:
I fixed some issues with this and pushed it.  Thanks!

PR #122, opened 2025-02-20
fix lldpad netlink heap access overrun

chunjiez:
If lldpad runs on top of some new versions of linux kernel (e.g. 6.10), then netlink message size would be larger than NLMSG_SIZE. In this case, lldpad application would crash due to heap access overrun.

### DIFF
--- a/lldp_util.c
+++ b/lldp_util.c
@@ -672,8 +672,9 @@ static struct nla_policy ifla_info_policy[IFLA_INFO_MAX + 1] =
 
 int is_macvtap(const char *ifname)
 {
-	int ret, s;
+	int ret, s, realsize;
 	struct nlmsghdr *nlh;
+	void *temp;
 	struct ifinfomsg *ifinfo;
 	struct nlattr *tb[IFLA_MAX+1],
 		      *tb2[IFLA_INFO_MAX+1];
@@ -684,13 +685,11 @@ int is_macvtap(const char *ifname)
 		return false;
 	}
 
-	nlh = malloc(NLMSG_SIZE);
+	nlh = calloc(1, NLMSG_SIZE);
 
 	if (!nlh) {
 		goto out;
 	}
-
-	memset(nlh, 0, NLMSG_SIZE);
 
 	nlh->nlmsg_len = NLMSG_LENGTH(sizeof(struct ifinfomsg));
         nlh->nlmsg_type = RTM_GETLINK;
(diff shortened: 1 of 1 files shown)


lldp_util.c
```diff
@@ -706,10 +704,23 @@ int is_macvtap(const char *ifname)
...
+	do {
+		realsize = recv(s, NULL, 0, MSG_DONTWAIT | MSG_PEEK | MSG_TRUNC);
+	} while ((realsize < 0) && errno == EINTR);
+
+	if (realsize <= 0) {
+		goto out_free;
+	}
+
+	free(nlh);
+	nlh = calloc(1, realsize);
```
penguin359: A `realloc()` and `memset()` might be better here. For the 99% of cases, this will be shrinking the buffer size and `realloc()` avoids having to move it around.

lldp_util.c
```diff
@@ -706,10 +704,23 @@ int is_macvtap(const char *ifname)
 		goto out_free;
 	}
 
-	memset(nlh, 0, NLMSG_SIZE);
+	do {
+		realsize = recv(s, NULL, 0, MSG_DONTWAIT | MSG_PEEK | MSG_TRUNC);
+	} while ((realsize < 0) && errno == EINTR);
+
+	if (realsize <= 0) {
```
penguin359: This should probably be `< 0`, otherwise it will never consume a zero-sized packet which is legitimate for some protocols like UDP, although I can't imagine netlink using that for anything. However, it is probably better to attempt to consume it if we are being pedantic here.

penguin359 (changes requested):
There should be no heap overrun or crash on the call to `recv()` because we pass the exact size of the allocated buffer. However, the parsing code of the netlink might have that possibility if they have embedded lengths and a truncated packet comes in. I'm not too familiar with how the netlink interface handles that case.

The logic generally seems sound with just a few comments added.

chunjiez:
@penguin359 , could you please review the updated commit?

penguin359 (approved):
I have finished my review and this seems reasonable. There is a possibility with the original code that if additional attributes are added to the message that the parsing code will overflow the default buffer size. From my own experiments with `rtnetlink(2)` this week-end, some netlink messages are indeed larger than 1024 bytes, but it seems none of the ones received by us so far.

It actually brings up a bigger issue with our use of the `libnl` API. We should be calling [`nlmsg_ok()`](https://www.infradead.org/~tgr/libnl/doc/api/group__msg.html#ga0107484894e5be506c1e1bb6cb09dbe2) on all received packets before any call to [`nlmsg_parse()`](https://www.infradead.org/~tgr/libnl/doc/api/group__msg.html#ga2671eefe95c3e44dec8ddab51ae57593) as that does the bounds checking to make sure that the buffer has enough bytes in it to account for the total size of the message listed in it's header. If we had that check, then we would simply return an error with the original code instead of crashing due to out-of-bounds memory access.

Also, `libnl` does offer a function to handle much of the logic that is being added here in [`nl_recv()`](https://www.infradead.org/~tgr/libnl/doc/api/group__send__recv.html#ga48ab5192dfa62d6b71760d43a4306d13) and friends, but are perhaps overkill with what they provide.

I see several other places in the code that seem to go into parsing netlink packets without a proper check that the advertised length is less than or equal to the buffer length for the packet. This includes a second, nearly identical use of `nlmsg_parse()` in this very same file. I approve this PR as is, but we should probably open an issue that tracks further improvements.

apconole:
Can you please squash into a single commit @chunjiez

chunjiez:
@apconole   done

apconole:
Thanks.  Applied, and added a fixes tag.

PR #126, opened 2025-11-17
lldp: fix small c&p mistakes

m-hau:
Fixes two small copy&paste mistakes.

<hr>

And a related piece of information for future researching users: To change the `sysName` TLV value (for example to the FQDN), you can use `lldptool -i eth0 -T -V sysName info=myhost.example.org`. This config option is unfortunately not listed by `lldptool -i eth0 -t -V sysName -c` until you set it.


penguin359:
Looks reasonable, but I want to refresh myself on this code before marking it approved.

apconole:
Merged, thanks!

